PR #15, opened 2025-03-02
Use use-cloud-event-proxy APIv2 only

rdavid:


test/conformance/serial/ptp.go
```diff
@@ -604,22 +604,6 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 				}
 				Expect(cloudProxyFound).ToNot(BeFalse(), "No event pods detected")
 
-				By("Checking event api is healthy")
```
jzding, line 607: do not remove these. They are needed to test v1 in older release

jzding (approved):
/lgtm

PR #20, opened 2025-03-12
Add telecomClockType to PtpSettings

josephdrichard:


josephdrichard:
Going to do this automatically instead of through a configurable field.

PR #21, opened 2025-03-17
Deprecation notice

rdavid:
If events is enabled and APIv1 is still in use after the upgrade to 4.19, the deprecation notice should appear in logs.


aneeshkp (approved):
/lgtm

PR #46, opened 2025-05-09
Fix reliability issue with CI

edcdavid:
If any command fails when creating certificates, fails fix-certs.sh so it can be retried.


edcdavid:
Checked in with other PR

PR #50, opened 2025-05-13
Added additional option to ExecCommand to be able to work in TTY

edcdavid:
Follow up to https://github.com/k8snetworkplumbingwg/ptp-operator/pull/49
Instead of moving to not using TTY, provide an option.


jzding (approved):
/lgtm

PR #60, opened 2025-06-02
Remove expect from promtheus metric check eventually

nocturnalastro:
As expect will call the fail handler it will cause the eventually negate the eventually.


jzding (approved):
/lgtm

PR #65, opened 2025-06-20
Remove NetworkPolicy for linuxptp-daemon

aneeshkp:
This PR removes the existing NetworkPolicy for the linuxptp-daemon pod.

Reason:

The linuxptp-daemon pod is configured with hostNetwork: true, meaning it shares the network namespace of the host node. Kubernetes NetworkPolicy resources do not apply to pods running with host networking, making the policy ineffective and unnecessary. Maintaining such a policy could lead to confusion or false assumptions about network isolation.


nocturnalastro (approved):
/lgtm

PR #67, opened 2025-07-02
Fix for failing downstream tests

edcdavid:
This PR includes 2 fixes:
- proper version check for readiness probe. previous library required a prefixed "v", So readiness probes were always checked even in versions that did not support it
- prometheus test failed because service needs authentication to work in Openshift. Rolled back changes and using exec into the prometheus pod directly.


test/pkg/ptphelper/ptphelper.go
```diff
@@ -435,8 +435,9 @@ func WaitForPtpDaemonToBeReady(podList []*v1core.Pod) int {
 		if err != nil {
 			isOCP = false
 		}
-
-		if isOCP && semver.Compare(ocpVersion, "4.19") < 0 {
+		ocpVer, _ := semver.NewVersion(ocpVersion)
```
nocturnalastro, line 438: I think this should actually be PTP operator version instead of OCP version as we don't redeploy the cluster each time now.

test/pkg/ptphelper/ptphelper.go
```diff
@@ -598,6 +606,34 @@ func GetPtpOperatorVersion() (string, error) {
 	return ptpOperatorVersion, err
 }
 
+func GetPtpOperatorVersionFromImage() (string, error) {
```
nocturnalastro, line 609: Might be better to get it from the BUILD_VERSION or RELEASE_VERSION in the env of the container.

PR #86, opened 2025-08-21
CNF-19145 ci: fix timeout onPushInitialEvent

jzding:
With fix for OCPBUGS-60570, the initial os-clock-sync-state-change event can only be pulled successfully when there is real data in the ptp event socket. Increase the timeout of the test case to make sure the data has been retrieved from the socket when check for initial event.


aneeshkp (approved):
/lgtm

PR #99, opened 2025-09-25
PTP-CI: Wait until the clock is in sync before checking initial roles

edcdavid:
Initial clock roles are sometime captured incorrectly in downstream CI because the clock takes more time to sychronize


nocturnalastro (approved):
/lgtm

PR #106, opened 2025-10-03
PTP Status Reporting via Custom Resource Conditions

aneeshkp:
PTP: add PtpConfig status conditions API, helper, and RBAC for operand updates
Introduces status.ptpStatus.conditions[] on PtpConfig to expose runtime health/state per profile and active config file.
Adds a small helper API for operands (linuxptp-daemon, cloud-event-proxy) to upsert conditions via the status subresource.
Wires RBAC so linuxptp-daemon can update PtpConfig status.
Documents the schema and examples in README.

> cond := ptpv1.PtpCondition{
>   Type:           ptpv1.ConditionPTP4lRunning,
>   Profile:        "GM-Profile-A",
>   Filename:       "ptp4l.0.config",
>   Status:         corev1.ConditionTrue,
>   Reason:         "Started",
>   Message:        "ptp4l is running",
>   LastUpdateTime: metav1.Now(),
> }
> _ = ptpstatus.UpsertPtpCondition(ctx, cs, "openshift-ptp", "default", cond)


pkg/status/ptpconfig_status.go
```diff
@@ -0,0 +1,52 @@
...
+            }
+        }
+        if !updated {
+            pc.Status.PtpStatus.Conditions = append(pc.Status.PtpStatus.Conditions, cond)
+        }
+
+        // enforce max history
+        const maxConditions = 100
+        if n := len(pc.Status.PtpStatus.Conditions); n > maxConditions {
+            pc.Status.PtpStatus.Conditions = pc.Status.PtpStatus.Conditions[n-maxConditions:]
```
josephdrichard, line 40: So if one process (e.g. ptp4l) has too many statuses, we won't see detail from other processes such as ts2phc if it didn't report more recently than the last 100 ptp4l?

ptp_real_failure_detector.sh
```diff
@@ -0,0 +1,182 @@
...
+set -e
+
+# Real PTP Failure Detector - No Test Mode
+# Queries actual Prow/GCS for real CI failures
+
+OPENSHIFT_VERSION="${OPENSHIFT_VERSION:-main}"
+LOOKBACK_HOURS="${LOOKBACK_HOURS:-24}"
+START_TIME="${START_TIME:-$(date -u -d "${LOOKBACK_HOURS} hours ago" +%Y-%m-%dT%H:%M:%SZ)}"
+
+echo "🔍 Checking for REAL PTP test failures since: $START_TIME"
```
nocturnalastro, line 11: Did you mean to add this? 
Looks like this is from your AI issue creater

aneeshkp: oops i thought 
i had deleted that

pkg/status/ptpconfig_status.go
```diff
@@ -0,0 +1,52 @@
...
+                updated = true
+                break
+            }
+        }
+        if !updated {
+            pc.Status.PtpStatus.Conditions = append(pc.Status.PtpStatus.Conditions, cond)
+        }
+
+        // enforce max history
+        const maxConditions = 100
```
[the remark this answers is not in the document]
aneeshkp: hmm, I think we thought of replacing same kind , but same kind means we need to match with conditionType and other paramters file profile name and file name , to support multiple  profiles

PR #119, opened 2025-10-29
CI: Test for Dual NIC BC with HA for phc2sys

greyerof:
Added a new test case for the DualNICBC in HA mode. The test case implements this sequence:
- Turn down the slave interface in the primary BC.
- Wait for phc2sys logs that reflect the selection of the secondary BC's slave interface.
- Turn up the original slave interface in the primary BC.
- Wait for ph2sys logs that reflect the selection of the original slave interface of the primary BC.

In order for this test case to work, I've done some modifications in the test framework:

- Added a function to create a separate third config for DualNIC Boundary Clocks working in HA mode. This extra config has phc2sys config opts and haProfiles in the ptpSettings field.
- Updated the discovery mechanism to detect the DualNICBC HA mode when two BC configs and the extra HA config for ph2sys are found.
- Refactored GetPodLogsRegex and created GetPodLogsRegexSince and moved the regex detection to a separate func findRegexInStream().

### DIFF
--- a/.github/workflows/aws-ci.yaml
+++ b/.github/workflows/aws-ci.yaml
@@ -60,7 +60,7 @@ jobs:
         id: launch
         run: |
           INSTANCE_ID=$(aws ec2 run-instances \
-            --image-id ami-0f7270e9a0e492eac \
+            --image-id ami-03f4e4724537a687c \
             --subnet-id subnet-04ee218fe8f32cad5 \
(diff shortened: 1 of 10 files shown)


edcdavid:
> @edcdavid I don't have permissions here to add labels. Could you add that ok-to-test one, please? Also, regarding the doc page, is it ok if I do that in a separate PR?

Sure !

PR #128, opened 2025-11-20
Allow additional behavior sources

vitus133:
This commit allows a new source type by making board label options in the behavior source configuration

/cc @nocturnalastro @lack


nocturnalastro (approved):
/approve

PR #135, opened 2025-12-04
Add clock type to hardwareconfig

vitus133:
This commit adds optional clock type to the hardwareconfig API and makes DPLL and Ethernet parts of the clock chain subsystem optional. This allows simplification of the user interface, where all the behavior details can be derived from ptpConfig and predefined hardware vendor details. For example, the T-BC hardware definition can be simplified as follows:
```
clockType: T-BC
clockChain:
      structure:
        - name: leader 
           # DPLL, sequences, labels, and Ports will be derived from ptpconfig 
           hardwareSpecificDefinitions: intel/e825 
 relatedPtpProfileName: 01-tbc-tr
```
The ability to override the defaults and configure flexible topologies will be preserved

/cc @josephdrichard @lack @nocturnalastro


nocturnalastro (approved):
/lgtm

PR #142, opened 2025-12-18
Add PTP Event API Version Validation Tests

yaronh12:
With v1 event API reaching EOL, we need to ensure:
- The cloud-event-proxy correctly defaults to v2 when no version is specified
- Explicit v2 configuration works as expected
- Invalid versions (v1 and garbage values) are properly rejected by the webhook

Added 4 new test specs to the "Running with event enabled" context:
1. **Default v2 behavior** - Verifies cloud-event-proxy runs v2.0 when `apiVersion` is not explicitly set in PtpOperatorConfig
2. **Explicit v2 setting** - Verifies cloud-event-proxy runs v2.0 when `apiVersion` is explicitly set to "2.0"
3. **v1 rejection** - Verifies setting `apiVersion` to "1.0" is rejected with EOL deprecation notice, config unchanged, pods not restarted
4. **Invalid version rejection** - Verifies setting `apiVersion` to invalid value ("boo") is rejected, config unchanged, pods not restarted

- Tests check cloud-event-proxy container logs for `REST API config: version=` to verify the running version
- Rejection tests verify webhook returns expected error message and confirm pods weren't restarted by comparing UIDs

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -1103,36 +1103,21 @@ spp 0
 
 					logrus.Infof("Checking cloud-event-proxy logs on pod %s (node: %s)", pod.Name, pod.Spec.NodeName)
 
-					// Get existing logs (no streaming)
-					podLogOptions := v1core.PodLogOptions{
-						Container: pkg.EventProxyContainerName,
-						Follow:    false,
-					}
-
-					req := client.Client.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &podLogOptions)
-					logs, err := req.DoRaw(context.Background())
+					// Search for REST API config v2.0 in pod logs (literal text search)
+					matches, err := pods.GetPodLogsRegex(
+						pod.Namespace,
+						pod.Name,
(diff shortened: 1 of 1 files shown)


test/conformance/serial/ptp.go
```diff
@@ -1075,6 +1075,201 @@ spp 0
...
+					Expect(err).NotTo(HaveOccurred(),
+						fmt.Sprintf("Failed to get logs from pod %s", pod.Name))
+
+					logStr := string(logs)
+					apiConfigPrefix := "REST API config: version="
+					lastIndexOfApiConfigSubstringInLogs := strings.LastIndex(logStr, apiConfigPrefix)
+					if lastIndexOfApiConfigSubstringInLogs == -1 {
+						Fail(fmt.Sprintf("No 'REST API config:' found in cloud-event-proxy logs on pod %s", pod.Name))
+					}
+					// Extract the version from the last occurrence
```
edcdavid: Why not use GetPodLogsRegex here instead of the code above? Same for similar code in this PR.

PR #151, opened 2026-01-04
test: add MITM attack resilience test and refactor security secrets (CNF-19576)

gtannous-spec:
## Summary

Add a functional test for MITM (Man-in-the-Middle) attack resilience when TLV Authentication is enabled, and refactor security-related test infrastructure for better maintainability. [CNF-19576](https://issues.redhat.com/browse/CNF-19576)

---

## Files Changed

| File | Change |
|------|--------|
| `test/conformance/serial/ptp.go` | Added MITM attack test case; refactored existing security tests to use shared helpers |
| `test/pkg/testconfig/testconfig.go` | Added secret creation helpers (`CreateMITMAttackerSecret`, `CreateMismatchSecret`) and secret name constants |
| `test/pkg/consts.go` | Added `sa_file` path constants for PTP security configuration |

---

## Test Added

**"PTP rejects MITM attack with same Key ID but different secret"**

This test validates that PTP authentication correctly rejects tampered packets:
1. Verifies baseline sync is healthy with proper authentication
2. Creates an "attacker" secret using the same Key IDs (1, 2) but different secret values
3. Switches the GM to use the attacker secret (simulating a MITM injecting packets)
4. Asserts the slave **fails to sync** for 2 minutes due to ICV (Integrity Check Value) mismatch
5. Restores original configuration and verifies recovery

---

## Refactoring: Magic Values → Constants

Moved inline secret definitions and hardcoded `sa_file` paths to shared locations:

- **Secret names** → `testconfig.MITMAttackerSecretName`, `testconfig.MismatchSecretName`
- **Secret creation** → `testconfig.CreateMITMAttackerSecret()`, `testconfig.CreateMismatchSecret()`
- **sa_file paths** → `pkg.SaFileSecurityConf`, `pkg.SaFileMismatchConf`, `pkg.SaFileMITMAttackerConf`

This improves readability, reduces duplication, and makes future security tests easier to implement.

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -561,27 +561,12 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 					// Delete if already exists
 					client.Client.Secrets(pkg.PtpLinuxDaemonNamespace).Delete(
 						context.Background(),
-						"ptp-security-mismatch",
+						testconfig.MismatchSecretName,
 						metav1.DeleteOptions{},
 					)
 					time.Sleep(2 * time.Second)
 
-					// Create secret with spp 0 but DIFFERENT KEYS than ptp-security-conf
-					// GM signs with these keys, BC has different keys for spp 0 → signature mismatch
-					mismatchSecret := &v1core.Secret{
-						ObjectMeta: metav1.ObjectMeta{
-							Name:      "ptp-security-mismatch",
-							Namespace: pkg.PtpLinuxDaemonNamespace,
-						},
-						StringData: map[string]string{
-							"ptp-security.conf": `[security_association]
-spp 0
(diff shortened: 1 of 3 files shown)


edcdavid (commented):
This does not look like a man-in-the-middle scenario, more like a misconfigured/rogue endpoint. The man in the middle imo should test:
- modifying all frames sent by the grandmaster for instance and make sure sync fails, or look for bad packet log after injecting 1 bad frame ->somewhat equivalent to proposed test but not exactly.
-  save and replay a valid frame later and monitor the log for 1 bad frame -> not covered

gtannous-spec (commented):
Thanks for the comment David, I actually missed the second point in the test. 
I will fix it, and then push another commit :)

gtannous-spec:
PTP CI negative test is moved to the PTP tech debt, will open a new PR for this issue later.

PR #171, opened 2026-01-31
CNF-19588: CI test for clockClass recovery after cloud-event-proxy crash

jzding:
## Summary
- Add e2e test verifying clock class metrics are recovered via the Event API after cloud-event-proxy crash and restart
- Portable process kill using `/proc/PID/comm` matching (works in both Kind/netdevsim and OCP hostPID environments)
- Recovery detection via health endpoint polling — handles both container restart (Kind) and process-level restart (OCP)
- Fix v2 API validation test: replace unreliable log search with direct health endpoint check
- Fix publisher readiness check: use explicit boolean flag instead of racy `time.Now()` comparison

## Changes
- `test/conformance/serial/ptp.go`: new crash recovery test (`Should recover clockClass via event API after cloud-event-proxy crash`), fixed v2 validation test, fixed publisher readiness check
- `test/pkg/event/event.go`: `WaitForCloudEventProxyReady`, `WaitForConsumerReady`, `PushInitialEvent`
- `test/pkg/ptphelper/ptphelper.go`: `PushInitialEvent` with context timeout

## Test plan
- [x] Passes on OCP 4.22 (cnfdg32, T-GM mode, real E810 hardware)
- [x] Passes on local Kind/netdevsim VM (BC mode)
- [x] CI: oc, bc, dualnicbc, dualnicbcha, dualfollower modes

Fixes: [CNF-19588](https://redhat.atlassian.net/browse/CNF-19588)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -64,6 +64,9 @@ const (
 var (
 	clockClassPattern = `^openshift_ptp_clock_class\{(?:config="ptp4l\.\d+\.config",)?node="([^"]+)",process="([^"]+)"\}\s+(\d+)`
 	clockClassRe      = regexp.MustCompile(clockClassPattern)
+
+	// pmcClockClassRe parses gm.ClockClass from "pmc GET PARENT_DATA_SET" output
+	pmcClockClassRe = regexp.MustCompile(`gm\.ClockClass\s+(\d+)`)
 )
 var DesiredMode = testconfig.GetDesiredConfig(true).PtpModeDesired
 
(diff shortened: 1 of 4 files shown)


test/conformance/serial/ptp.go
```diff
@@ -3166,6 +3268,30 @@ func checkClockClassState(fullConfig testconfig.TestConfig, expectedState string
 		fmt.Sprintf("Expected ptp4l clock class to eventually be %s for GM", expectedState))
 }
 
+// checkClockClassViaPMC verifies clock class by running "pmc GET PARENT_DATA_SET"
+// and parsing the gm.ClockClass field from the output.
+func checkClockClassViaPMC(fullConfig testconfig.TestConfig, expectedClockClass string) {
```
nocturnalastro: I think VerifyNICClockClass already does this no?

test/pkg/ptphelper/ptphelper.go
```diff
@@ -1037,7 +1037,7 @@ func IsPTPOperatorVersionAtLeast(minVersion string) bool {
 		logrus.Infof("Could not get PTP Operator version, assuming version check passes: %v", err)
 		return true
 	}
-	logrus.Infof("Found version %s; checking %s <= %s", foundVersion, foundVersion, minVersion)
+	logrus.Infof("Found version %s; checking if %s >= %s", foundVersion, foundVersion, minVersion)
```
nocturnalastro: It should be just `<` 
```go
	return !ver.LessThan(minVer)
```

test/pkg/testconfig/testconfig.go
```diff
@@ -1112,20 +1112,23 @@ func PtpConfigOC(isExtGM bool) error {
 
 	BestSolution := ""
 
-	if isExtGM {
```
edcdavid, line 1351: External grandmaster is a mode configured with EXTERNAL_GM=true in combination with the clock mode. For instance OC + external GM is a distinct scenario from OC + Local GM. By default it is false. The solver always reports all possible scenarios but the test should check only the scenario requested by the user. I would revert the changes in this file to let the user know that his requested mode is not available(local GM for instance).

test/conformance/serial/ptp.go
```diff
@@ -1359,6 +1359,105 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
...
 
+			It("Should recover clockClass via event API after cloud-event-proxy crash", func() {
+				if ptphelper.PtpEventEnabled() != 2 {
+					Skip("Skipping: test applies to event API v2 only")
+				}
+
+				// Determine expected clock class based on PTP mode
+				var expectedClockClass fbprotocol.ClockClass
+				switch fullConfig.PtpModeDiscovered {
+				case testconfig.OrdinaryClock:
```
nocturnalastro: What clock type wouldn't be 6 via PMC when locked?

I think this should just be an 

```go
expectedClockClass := fbprotocol.ClockClass6
if fullConfig.PtpModeDiscovered == testconfig.OrdinaryClock && ptphelper.IsPTPOperatorVersionAtLeast("4.21") {
    expectedClockClass = fbprotocol.ClockClassSlaveOnly
}
```

test/conformance/serial/ptp.go
```diff
@@ -1359,6 +1359,105 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
...
+						pkg.EventProxyContainerName,
+						[]string{"sh", "-c", "pgrep -f ^./cloud-event-proxy"},
+					)
+					return strings.TrimSpace(buf.String()) != ""
+				}, 3*time.Minute, 1*time.Second).Should(BeTrue(),
+					"cloud-event-proxy process did not restart within 3 minutes")
+
+				By("Waiting for cloud-event-proxy health endpoint to recover")
+				Eventually(func() string {
+					buf, _, _ := pods.ExecCommand(
```
nocturnalastro: Seems like this should just be a helper something like WaitForCloudEventProxyToBeReady?

nocturnalastro: or maybe there should be an entire one for KillCloudEventProxy

jzding: will create a helper WaitForCloudEventProxyReady(). It is more likely to be reused then a helper for KillCloudEventProxy

test/conformance/serial/ptp.go
jzding: good point , updated

nocturnalastro (approved):
/lgtm

PR #196, opened 2026-03-16
feat : GNSS Simulation

bnshr:
## Summary

Adds a software GNSS simulator to enable T-GM (Telco Grand Master) end-to-end testing in CI without physical GNSS receivers, Intel WPC NICs, or hardware DPLL.

## Motivation / Problem

[JIRA Ticket](https://redhat.atlassian.net/browse/CNF-21632)

The existing T-GM test suite (`PTP_TEST_MODE=tgm`) requires hardware that is unavailable in the Kind-cluster-based CI environment:

- Intel E810 WPC NICs (subsystem device `000e`)
- Physical GNSS receivers at `/dev/gnss`
- Hardware DPLL for clock synchronization
- u-blox GPS modules controlled via `ubxtool`

This blocks upstream CI coverage for T-GM clock state verification, holdover scenarios, and PTP event API validation.

## Solution

A new `ptp-tools/gnss-sim/` Go binary generates valid NMEA sentences (GNRMC, GNGGA, GPZDA) at 1 PPS over socat virtual serial ports. It includes:

- **DPLL state machine** — models LOCKED → HOLDOVER → FREERUN with clock class transitions (CC6 → CC7 → CC248)
- **HTTP control API** — `/api/signal/loss` and `/api/signal/restore` replace `ubxtool` commands for test orchestration
- **socat PTY integration** — `ts2phc` reads from `/dev/ttyGNSS_TS2PHC` as if it were a real GNSS device

A new test mode `PTP_TEST_MODE=tgm-sim` activates the simulation. WPC NIC detection functions are mocked at the test helper level so **`CreatePtpConfigWPCGrandMaster()` — the real config creation path — runs unchanged**. The only conditional is skipping the E810 plugin (pin maps, ubxtool) since netdevsim has no Intel ICE driver.

## Technical Details

**Architecture decision — mock detection, not config creation:**
Rather than creating a separate sim-specific config function, the detection layer (`getWPCEnabledIfaces`, `checkGNSSAvailabilityForIface`) returns simulated data when `IsSimulatedTGM()` is true. This keeps the config creation path identical to production. L2 discovery is bypassed because netdevsim interfaces cannot satisfy the solver's `StepIsWPCNic` constraint.

**Key files:**

| Area | Files |
|------|-------|
| Simulator | `ptp-tools/gnss-sim/` — main, nmea, simulator, dpll, api, entrypoint, unit tests |
| Container | `ptp-tools/Dockerfile.gnss-sim`, `scripts/configGNSS.sh` |
| Detection mock | `test/pkg/ptphelper/ptphelper.go` |
| Config path | `test/pkg/testconfig/testconfig.go` |
| Conformance tests | `test/conformance/serial/ptp.go` — 3 new contexts (verification, signal loss, CloudEvents v2) |
| CI pipeline | `scripts/run-on-vm.sh`, `scripts/run-tests.sh`, `ptp-tools/Makefile` |

**Breaking changes:** None. All changes are additive; existing test modes are unaffected.

**New dependencies:** None. The simulator is a standalone Go binary; the only runtime dependency is `socat` (installed in the container image).

## Testing

- **Unit tests:** 17/17 pass — NMEA generation, checksum integrity, DPLL state transitions, PPS status
- **`make fmt` + `go vet`:** clean
- **Ginkgo dry-run:** 5 simulated T-GM specs compile and are discoverable via `--focus=".*Simulated T-GM.*"`
- **Full e2e:** requires Kind cluster with netdevsim; run with `PTP_TEST_MODE=tgm-sim`

## Risks / Considerations

> **Reviewers should pay attention to:**

- **`ptphelper.go` detection mocks** — `getWPCEnabledIfaces()` and `checkGNSSAvailabilityForIface()` have early-return branches when `IsSimulatedTGM()`. Verify these don't affect real `tgm` mode (they are gated on `PTP_TEST_MODE=tgm-sim` only).
- **E810 plugin skip** — `CreatePtpConfigWPCGrandMaster()` passes `nil` plugins in sim mode. Confirm this doesn't break `createConfigWithTs2PhcAndPlugins()` when plugins are nil.
- **gpsd/gpspipe exclusion** — the simulated environment intentionally skips these processes since gnss-sim writes directly to the PTY. The sim-aware tests check only `ts2phc`, `phc2sys`, and `ptp4l`.

### Reviewer Checklist

- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage
- [ ] Sufficient end-to-end test coverage
- [ ] Bug fixes are accompanied by regression test(s)
- [ ] e2e tests and flake fixes are accompanied evidence of flake testing, e.g. executing the test 100(0) times
- [ ] tech debt/todo is accompanied by issue link(s) in comments in the surrounding code
- [ ] Tests are comprehensible, e.g. Ginkgo DSL is being used appropriately
- [ ] Docs updated or added to /doc
- [ ] Commit messages sensible and descriptive
- [ ] Tests marked as [FLAKE] are truly flaky and have an issue
- [ ] Code is properly formatted

Assisted-by AI tool


test/conformance/serial/ptp.go
```diff
@@ -2344,6 +2344,278 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 			})
 		})
 
+		Context("Simulated T-GM Verification Tests", func() {
+			BeforeEach(func() {
+				if !ptphelper.IsSimulatedTGM() {
+					Skip("test valid only for simulated T-GM (PTP_TEST_MODE=tgm-sim)")
```
edcdavid: We should not create a new tgm-sim, the existing tgm scenario should already cover it by discovering the WPC card that provides the gnss hardware. The simulator should appear to the script as real HW so that ptp-operator performs as if on real hardware. The virtual NIC supporting gnss should have a pci ID that will make it be discovered as gnrd (supporting gnss)

edcdavid (commented):
Added some comments. Please add a detailed PR description for the new feature. This feature should be integrated with l2 discovery so that any scenario using gnss or telecom GM can discover this gnss.

ptp-tools/gnss-sim/api.go
```diff
@@ -0,0 +1,180 @@
...
+	s.mux.HandleFunc("/api/status", s.handleStatus)
+	s.mux.HandleFunc("/api/dpll", s.handleDPLLStatus)
+	s.mux.HandleFunc("/api/signal/loss", s.handleSignalLoss)
+	s.mux.HandleFunc("/api/signal/restore", s.handleSignalRestore)
+	s.mux.HandleFunc("/api/config", s.handleConfig)
+	return s
+}
+
+// ListenAndServe starts the HTTP server on the given address (e.g. ":9200").
+func (s *APIServer) ListenAndServe(addr string) error {
```
gtannous-spec, line 36: are we referring to this function in our codebase?

I see it is called via `http` instance at the `return`, but the wrapper is never called somewhere.

test/conformance/serial/ptp.go
```diff
@@ -2722,6 +3009,14 @@ func checkStabilityOfWPCGMUsingMetrics(fullConfig testconfig.TestConfig) {
 	checkPTPNMEAStatus(fullConfig, "1")
 }
 
+// checkStabilityOfSimGMUsingMetrics checks simulated T-GM stability
+// using only the processes and metrics available in the simulated environment.
+func checkStabilityOfSimGMUsingMetrics(fullConfig testconfig.TestConfig) {
```
gtannous-spec, line 3014: will this be referenced in our codebase?

test/pkg/ptphelper/ptphelper.go
```diff
@@ -1043,3 +1095,253 @@ func IsPTPOperatorVersionAtLeast(minVersion string) bool {
...
+	return gnssSimNmeaActive.Load()
+}
+
+// IsGnssSimulatedCI is true for conformance paths that target gnss-sim instead of gpsd/hardware GNSS.
+func IsGnssSimulatedCI() bool {
+	return UseGnssSimulation()
+}
+
+// IsIntegratedGnssSim is an alias for UseGnssSimulation (historical name).
+func IsIntegratedGnssSim() bool {
```
gtannous-spec, line 1235: will this be referenced in the codebase?

test/pkg/consts.go
```diff
@@ -28,6 +28,7 @@ const (
 	// policy name
 	PtpGrandMasterPolicyName      = "test-grandmaster"
 	PtpWPCGrandMasterPolicyName   = "test-wpc-grandmaster"
+	PtpSimGrandMasterPolicyName   = "test-sim-grandmaster"
```
gtannous-spec, line 31: this const is used in cleanup only but never in testconfig for config creations

test/pkg/ptphelper/ptphelper.go
```diff
@@ -1043,3 +1095,253 @@ func IsPTPOperatorVersionAtLeast(minVersion string) bool {
...
+	if port == "" {
+		port = "9200"
+	}
+	return fmt.Sprintf("http://%s:%s", host, port)
+}
+
+// GNSSSimSignalLoss calls POST /api/signal/loss on the gnss-sim, equivalent
+// to ubxtool CFG-NAVSPG-INFIL_NCNOTHRS,50,1 on real hardware.
+func GNSSSimSignalLoss() error {
+	resp, err := http.Post(GNSSSimAPIBase()+"/api/signal/loss", "application/json", nil)
```
gtannous-spec, line 1131: this is using a http method without a timeout (unless the default handles a certain timeout). 

I see this is called in the test suite, in case there is a crash in the gnss-container level the `http calls` will be blocked causing the entire CI test suite to freeze without log message instead of retry failure. 

Would probably be worthy adding a http client instance with a reasonable timeout and replace all the bare calls with it. WDYT @edcdavid

test/pkg/ptphelper/ptphelper.go
```diff
@@ -1043,3 +1095,253 @@ func IsPTPOperatorVersionAtLeast(minVersion string) bool {
...
+		return fmt.Errorf("gnss-sim signal loss returned %d: %s", resp.StatusCode, string(body))
+	}
+	logrus.Info("gnss-sim: signal loss triggered")
+	return nil
+}
+
+// GNSSSimSignalRestore calls POST /api/signal/restore on the gnss-sim, equivalent
+// to ubxtool CFG-NAVSPG-INFIL_NCNOTHRS,0,1 on real hardware.
+func GNSSSimSignalRestore() error {
+	resp, err := http.Post(GNSSSimAPIBase()+"/api/signal/restore", "application/json", nil)
```
gtannous-spec, line 1147: same as above.

test/pkg/ptphelper/ptphelper.go
```diff
@@ -1043,3 +1095,253 @@ func IsPTPOperatorVersionAtLeast(minVersion string) bool {
...
+	State           string `json:"state"`
+	FrequencyStatus int    `json:"frequencyStatus"`
+	PhaseStatus     int    `json:"phaseStatus"`
+	ClockClass      int    `json:"clockClass"`
+	PPSStatus       int    `json:"ppsStatus"`
+}
+
+// GNSSSimGetDPLLState retrieves the current DPLL state from the gnss-sim API.
+func GNSSSimGetDPLLState() (*GNSSSimDPLLState, error) {
+	resp, err := http.Get(GNSSSimAPIBase() + "/api/dpll")
```
gtannous-spec, line 1171: same as above.

test/pkg/ptphelper/ptphelper.go
```diff
@@ -1043,3 +1095,253 @@ func IsPTPOperatorVersionAtLeast(minVersion string) bool {
...
+	SignalActive bool    `json:"signalActive"`
+	GPSFix       int     `json:"gpsFix"`
+	Satellites   int     `json:"satellites"`
+	HDOP         float64 `json:"hdop"`
+	OffsetNs     int64   `json:"offsetNs"`
+}
+
+// GNSSSimGetStatus retrieves the current simulator state from the gnss-sim API.
+func GNSSSimGetStatus() (*GNSSSimStatus, error) {
+	resp, err := http.Get(GNSSSimAPIBase() + "/api/status")
```
gtannous-spec, line 1198: same as above

test/pkg/ptphelper/ptphelper.go
```diff
@@ -1043,3 +1095,253 @@ func IsPTPOperatorVersionAtLeast(minVersion string) bool {
...
+	var status GNSSSimStatus
+	if err := json.NewDecoder(resp.Body).Decode(&status); err != nil {
+		return nil, fmt.Errorf("gnss-sim status decode failed: %v", err)
+	}
+	return &status, nil
+}
+
+// GNSSSimIsHealthy returns true if the gnss-sim /health endpoint responds with 200 OK.
+func GNSSSimIsHealthy() bool {
+	resp, err := http.Get(GNSSSimAPIBase() + "/health")
```
gtannous-spec, line 1216: same as above

ptp-tools/gnss-sim/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/k8snetworkplumbingwg/ptp-operator/ptp-tools/gnss-sim
+
+go 1.22
```
gtannous-spec, line 3: probably need to change this from `go 1.22` to `go 1.25` to align with the main project's Go version.

ptp-tools/gnss-sim/api.go
```diff
@@ -0,0 +1,180 @@
...
+//   - linuxptp-daemon: ts2phc reports nmea_status 0, servo s0
+//   - cloud-event-proxy: publishes GnssStateChange with FAILURE_NOFIX
+//   - DPLL: transitions LOCKED → HOLDOVER → FREERUN
+//   - Clock class: 6 → 7 → 248
+func (s *APIServer) handleSignalLoss(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+	s.state.SetSignal(false)
```
gtannous-spec, line 93: the signal loss mutation is not atomic across fields, each `SetSignal` and `SetGPSFix` acquire independently `SimState.mu` as I can see in `simulator.go` .
Between these two calls, the Go scheduler can run either of the two go-routines that concurrently read this same state object..
If either go-routine calls `Snapshot()` after `SetSignal(false) `returns but before `SetGPSFix(GPSFixNoFix) `executes, the snapshot will contain `SignalActive=false` with `GPSFix=3D`  a contradictory state that real GNSS hardware can't handle.

It's not guaranteed to happen as it depends on when the scheduler runs each go-routine. But it can happen, and in a test environment it might cause into a race condition.. it is time dependent and non-deterministic.

I guess it would be better to wrap this 2 correlated fields update under a single mutex lock. WDYT @edcdavid

ptp-tools/gnss-sim/api.go
```diff
@@ -0,0 +1,180 @@
...
+//
+//	ubxtool -P 29.25 -w 1 -v 3 -z CFG-NAVSPG-INFIL_NCNOTHRS,0,1
+//
+// Sets signal active with 3D fix and 12 satellites.
+func (s *APIServer) handleSignalRestore(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+	s.state.SetSignal(true)
```
gtannous-spec, line 112: same as above for these 3 lines

PR #199, opened 2026-03-18
CNF-22218: Clean CI with multi level logs

shirmoran:
- Added date/time info to the logs.
- cleaned log info level - moved most of the logs to the debug level.
- Added relevant information with Ginkgo "By" statements.


edcdavid (commented):
Few comments:
- Also add timestamps on log lines generated from the run-on-vm.sh script, add color for steps.
- When building images using ptp-tools scripts, hide podman logs by default. Only display start and end of image building and pushing. Use color to highlight these steps. If any strep fails, then dump the build/push logs leading to the failure.
- Actually, reviewing the verbosity vs use of logs, most seem ok currently to me. I would revert most of the info -> debug changes and look at the few repetitive/verbose ones that repeat info (like waiting for an state, dumping non human readable info -> reformat?). 
- Might be useful to clarify logs that are not clear, if needed

test/conformance/parallel/parallel_suite_test.go
```diff
@@ -64,26 +68,34 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
-		}
-	} else {
-		logrus.Info("Deploy consumer app without sidecar for testing event API v2")
-		err = event.CreateConsumerApp(fullConfig.DiscoveredClockUnderTestPod.Spec.NodeName)
-		if err != nil {
-			logrus.Errorf("PTP events are not available due to consumer app creation error err=%s", err)
-			isConsumerReady = false
+
+	By("Enabling PTP event API", func() {
+		logrus.Info("API version %s", apiVersion)
```
gtannous-spec, line 73: `logrus.Info`  used with format string will not interpolate, it doesn't interpret the %s. 

This should be `logrus.Infof ` or `logrus.Debugf` to match the log-level reduction.


```suggestion
		logrus.Debugf("API version %s", apiVersion) 
```

scripts/run-on-vm.sh
```diff
@@ -1,31 +1,82 @@
...
+  local phase="$2"
+  local phase_label="${phase_type} ${phase}"
+  shift
+  shift
+
+  local log_file
+  log_file="$(mktemp "/tmp/ptp-operator/${phase_label// /_}.XXXXXX.log")"
+
+  echo -e "${COLOR_PHASE}START ${phase_type}:${COLOR_RESET} ${phase}"
+  if "$@" >"${log_file}" 2>&1; then
```
gtannous-spec, line 38: This function always returns 0 on failure after this `if` fails, then `rc=$?` no longer holds the command's exit code. It merely holds the exit status of the `if` compound statement itself, which is always 0 when there's no `else` branch. So `local rc=$?` captures 0, the log says "exit code 0", and `return "${rc}"` tells the caller everything succeeded.

Since the script runs under `set -e`, this means failed `make podman-buildall` , `make podman-pushall` calls will be silently ignored and the script will continue deploying with missing or stale images.

would better be moving the failure path into an `else` branch, where `$?` still holds the original exit code.

scripts/run-on-vm.sh
```diff
@@ -1,31 +1,82 @@
...
-set -x
 set -euo pipefail
+set -x
+
+# Save full run output under /tmp/ptp-operator (timestamped file; also shown on the terminal).
+mkdir -p /tmp/ptp-operator
+RUN_ON_VM_LOG="/tmp/ptp-operator/run-on-vm-$(date +%Y%m%d-%H%M%S).log"
+
+# Prefix all script output lines with timestamp, then tee to RUN_ON_VM_LOG.
+exec > >(awk '{ print strftime("%Y-%m-%d %H:%M:%S"), $0; fflush(); }' | tee "$RUN_ON_VM_LOG") 2>&1
```
gtannous-spec, line 10: Eliminate double-timestamping pick either `PS4` or `awk`, not both. 

in fact `awk` can handle everything uniformly.

test/pkg/testconfig/testconfig.go
```diff
@@ -440,7 +475,7 @@ func initFoundSolutions() {
 
 // Gets te desired configuration from the environment
 func GetDesiredConfig(forceUpdate bool) TestConfig {
-	defer logrus.Infof("Current PTP test config=%s", &GlobalConfig)
+	defer logrus.Debugf("Current PTP test config:\n%s", reformatGlobalConfig(&GlobalConfig))
```
gtannous-spec, line 478: Be careful, all arguments to a deferred function are evaluated immediately when the defer statement is reached, not when the surrounding function returns. 
`reformatGlobalConfig(&GlobalConfig)` runs immediately, marshals the config to JSON, and stores that string. When the function returns, logrus.Debugf is called with the stale, pre-modification snapshot not the final state.

test/pkg/testconfig/testconfig.go
```diff
@@ -749,8 +784,8 @@ func initAndSolveProblems() {
 
 // Gets the discovered configuration
 func GetFullDiscoveredConfig(namespace string, forceUpdate bool) TestConfig {
-	logrus.Infof("Getting ptp configuration for namespace:%s", namespace)
-	defer logrus.Infof("Current PTP test config=%s", &GlobalConfig)
+	logrus.Debugf("Getting ptp configuration for namespace:%s", namespace)
+	defer logrus.Debugf("Current PTP test config:\n%s", reformatGlobalConfig(&GlobalConfig))
```
gtannous-spec, line 788: same as above.

PR #210, opened 2026-04-13
CNF-22878: Fix VerifyNICClockClass to handle missing config label on OCP < 4.18

aabughosh:
## Summary

- **Fixes version-dependent CI failures** where `VerifyNICClockClass` timed out waiting for the `openshift_ptp_clock_class{config=...}` per-config Prometheus metric on OCP 4.16/4.17. The `config` label was only added to this metric in `cloud-event-proxy` starting from OCP 4.18.
- **Adds version-aware metric verification** using `IsPTPOperatorVersionAtLeast("4.18")`:
  - **OCP >= 4.18**: PMC + per-config metric (with `config` label)
  - **OCP < 4.18, single NIC**: PMC + node-level metric (without `config` label)
  - **OCP < 4.18, dual NIC**: PMC only (a single node-level metric cannot distinguish between two NICs)
- **Passes `isDualNIC` parameter** to `VerifyNICClockClass` from `ptp.go` so the helper knows which verification path to take.

## Root Cause

On OCP 4.16/4.17, `cloud-event-proxy` registers `openshift_ptp_clock_class` with only `node` and `process` labels (no `config` label). The existing `perConfigClockClassRe` regex expects `config="..."`, so it never matched, causing 5-minute timeouts on every call to `VerifyPerConfigClockClassWithMetrics`.

## Changes

| File | Change |
|------|--------|
| `test/pkg/ptptesthelper/ptptesthelper.go` | Added `nodeClockClassRe`, `getNodeClockClassFromMetrics`, `verifyNodeClockClassWithMetrics`; refactored `VerifyNICClockClass` to be version- and NIC-aware |
| `test/conformance/serial/ptp.go` | Pass `isDualNIC` (`false` for single NIC, `true` for dual NIC) to all `VerifyNICClockClass` calls |

## Linked Issues

- [CNF-22878](https://redhat.atlassian.net/browse/CNF-22878)

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -1111,21 +1111,21 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 
 				// Step 1: verify initial locked state via PMC, metrics, and events
 				By("Step 1: Verifying NIC reports clockClass 6 (initial locked state)")
-				ptptesthelper.VerifyNICClockClass(fullConfig, nic1, locked)
+				ptptesthelper.VerifyNICClockClass(fullConfig, nic1, locked, false)
 				verifyClockClassViaEvent(evCtx, locked)
 
 				// Step 2: cut upstream sync by bringing slave interface down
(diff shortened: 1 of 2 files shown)


edcdavid (dismissed):
Looks good to me after gemini comment fix

edcdavid (approved):

PR #211, opened 2026-04-14
Ci test helper improvements

nocturnalastro:


test/conformance/serial/ptp.go
```diff
@@ -274,8 +275,18 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 				Skip("Skipping as the test is only applicable for Dual NIC BC in HA mode (dualnicbcha)")
 			}
 
+			By("Checking if HA is availble in this opertator version")
+			ptpOperatorVersion, err := ptphelper.GetPtpOperatorVersion()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ptpOperatorVersion).ShouldNot(BeEmpty())
+			operatorVersion, _ := semver.NewVersion(ptpOperatorVersion)
+			haVersion, _ := semver.NewVersion("4.16")
```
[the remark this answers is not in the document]
nocturnalastro: 4.16 is a valid version there is no need to check the error. PTP version should also be but we'll check that one.

test/pkg/ptphelper/ptphelper.go
```diff
@@ -899,21 +899,65 @@ func GetListOfWPCEnabledInterfaces(nodeName string) ([]string, string) {
...
+				return group, deviceId
 			}
 		}
 	}
+
 	return retList, deviceId
 }
-func addAllInterfacesForNic(WPCifaces map[string]string, firstIface string) []string {
+
+func addAllInterfacesForNic(nodeName string, WPCifaces map[string]string, firstIface string) []string {
```
sebsoto: I know this is existing code, but since its being extensively changed... 😬 

Is `addAllInterfacesForNic` adding to anything? Not sure the name makes sense.
Could it be renamed to `getInterfacesForNic`? 
(Dropped `all` too because that is implied) 

The significance of `firstIface` isn't clear. First relevant to what?
Does this function only work for WPC? Asking because of the argument `WPCifaces`. If so that should be reflected in someway.

test/pkg/ptphelper/ptphelper.go
```diff
@@ -942,6 +986,56 @@ func getWPCEnabledIfaces(nodeName string) map[string]string {
...
+			if parseErr != nil {
+				return -1, fmt.Errorf("unable to parse PTP Hardware Clock index from %q", line)
+			}
+			return idx, nil
+		}
+	}
+	return -1, fmt.Errorf("PTP Hardware Clock not found in ethtool output for %s", ifaceName)
+}
+
+func findIfaceWithPinsForPhc(nodeName string, phcIndex int) (string, error) {
```
sebsoto, line 1028: A doc for this function would be helpful.

test/pkg/ptphelper/ptphelper.go
sebsoto: Maybe rename to `GetWPCEnabledInterfaces` or just `WPCEnabledInterfaces`?

List of is implied by both the plural `Interfaces` and the []string return type.

sebsoto: Given this has a non-standard signature (returns []string,string), it would be really helpful to a function doc indicating the return values. The []string is obviously the interface list, the second string is hard for me to understand having skimmed the function. Is it the interface with GNSS?

test/pkg/ptphelper/ptphelper.go
```diff
@@ -899,21 +899,65 @@ func GetListOfWPCEnabledInterfaces(nodeName string) ([]string, string) {
...
+			return group[i] < group[j]
+		})
+		ifaceWithPins, err := findIfaceWithPinsForPhc(nodeName, idx)
+		if err != nil {
+			logrus.Debugf("PTP pins lookup failed phc=%d err=%v", idx, err)
+			continue
+		}
+		for _, iface := range group {
+			if iface == ifaceWithPins {
+				deviceId, _ = checkGNSSAvailabilityForIface(nodeName, iface)
```
sebsoto: The error is being swallowed here.
Given  this calls out to  
```	cmd := []string{"/bin/sh", "-c", fmt.Sprintf("ls /sys/class/net/%s/device/gnss", IfaceName)}```
Failure is definitely possible due to permissions and should be handled appropriately. I highly recommend either adding an error return  to the function, or ideally separating the interface gathering and the GNSS checking into distinct functions. This function is doing too much as it is, in my opinion.

File: test/pkg/ptphelper/ptphelper.go [hunk 5]
```diff
@@ -942,6 +986,56 @@ func getWPCEnabledIfaces(nodeName string) map[string]string {
 	return resMap
 }
 
+func getPTPHardwareClockIndex(nodeName string, ifaceName string) (int, error) {
+	cmd := []string{"ethtool", "-T", ifaceName}
+	so, se, err := execPodCommand(nodeName, cmd)
+	if err != nil {
+		return -1, fmt.Errorf("ethtool -T %s failed: %s", ifaceName, strings.TrimSpace(se.String()))
```
sebsoto: Why the -1 return? Makes sense from a program exit code, but not so much within a go program.

Failure should be captured in the functions error, callers having to check both the error and the return value is cumbersome.

test/conformance/serial/ptp.go
```diff
@@ -275,8 +275,19 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 				Skip("Skipping as the test is only applicable for Dual NIC BC in HA mode (dualnicbcha)")
 			}
 
+			By("Checking if HA is availble in this opertator version")
```
sebsoto: s/opertator/operator 🥔

sebsoto (changes requested):

sebsoto (approved):
/lgtm

PR #212, opened 2026-04-15
fix: add sync gates to eliminate CI test race conditions

gtannous-spec:
### This PR is mainly for testing whether we get more consistent tests passing and that we get rid of flaky fails.

Add synchronization gates and retry logic to the conformance test helpers to ensure tests wait for the full operator→configmap→daemon pipeline to complete before asserting on logs and metrics.

CI conformance tests are flaky because they proceed before the system has reached a consistent state. Log-based lookups fail when the daemon hasn't finished starting, and pod replacements race against operator reconciliation, causing non-deterministic failures that vary in location from run to run.

Tests now explicitly gate on each stage of the pipeline: the configmap reflecting the expected profile, the daemon logging that it loaded the profile, and the foreign master appearing in the log stream. This replaces the previous eager-fail behavior where a single transient log gap would abort the entire test before any retry logic could run.

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -907,11 +907,24 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 						}
 					})
 
+					// Wait for operator to reconcile the config map with the
+					// temp profile before restarting the daemon pod, so the new
+					// pod starts with the correct config on first boot.
+					err = ptphelper.WaitForConfigMapProfile(nodes.Items[0].Name, pkg.PtpTempPolicyName, 2*time.Minute)
+					Expect(err).NotTo(HaveOccurred(), "operator did not reconcile temp profile into configmap in time")
+
 					testPtpPod, err = ptphelper.GetPtpPodOnNode(nodes.Items[0].Name)
 					Expect(err).NotTo(HaveOccurred())
 
 					testPtpPod, err = ptphelper.ReplaceTestPod(&testPtpPod, time.Minute)
 					Expect(err).NotTo(HaveOccurred())
+				})
+
+				By("Waiting for daemon to load the temp profile", func() {
+					_, err := pods.GetPodLogsRegex(testPtpPod.Namespace,
+						testPtpPod.Name, pkg.PtpContainerName,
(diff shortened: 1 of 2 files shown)


nocturnalastro (dismissed):
/lgtm

test/pkg/ptptesthelper/ptptesthelper.go
edcdavid, line 56: Looks like this call is redundant, WaitForClockIDForeign already achieves the same thing. I would remove it.

gtannous-spec: I actually reverted the whole changes in this File for now just trying to pay attention to multiple scenarios where the tests can have flaky failures.

test/pkg/ptptesthelper/ptptesthelper.go
```diff
@@ -61,21 +61,24 @@ func BasicClockSyncCheck(fullConfig testconfig.TestConfig, ptpConfig *ptpv1.PtpC
...
+		// Use WaitForClockIDForeign to handle both cases:
+		// (a) GetClockIDForeign failed entirely (log not available yet)
+		// (b) slaveMaster doesn't match expected GM (stale entry or re-sync in progress)
+		// This avoids the eager-fail where a transient log absence aborts the test
+		// before the retry/metrics checks can run.
+		if err != nil || slaveMaster == "" || !strings.HasPrefix(slaveMaster, *gmID) {
+			logrus.Infof("slaveMaster=%q does not match expected GM=%s (err=%v), waiting for re-sync...", slaveMaster, *gmID, err)
 			if waitErr := ptphelper.WaitForClockIDForeign(profileName, label, nodeName, *gmID); waitErr != nil {
-				return errors.Errorf("Slave connected to another (incorrect) Master, slaveMaster=%s, gmID=%s, waitErr=%s", slaveMaster, *gmID, waitErr)
+				return errors.Errorf("Slave did not sync to expected Master, slaveMaster=%s, gmID=%s, waitErr=%s", slaveMaster, *gmID, waitErr)
```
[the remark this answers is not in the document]
gtannous-spec: irrelevant anymore

edcdavid (approved):

PR #220, opened 2026-04-27
test: add transport-level retry and harden K8s API resilience

edcdavid:
## Summary

- Add a `retryRoundTripper` via `rest.Config.Wrap()` that automatically retries transient HTTP errors (connection refused, i/o timeout, TLS handshake timeout, HTTP 429, 5xx) with exponential backoff for **every** K8s API call, replacing scattered per-call-site retry logic.
- Pre-wait up to 12 minutes for namespaces stuck in `Terminating` before calling into the `privileged-daemonset` library (whose own namespace delete timeout is only 2 minutes). This addresses CI failures observed in Prow dashboard data for three namespaces:
  - **`l2discovery`** — L2 discovery init via `testconfig.CreatePtpConfigurations`
  - **`ptp-network-outage-recovery`** — network outage recovery tests via `ptptesthelper.CreatePtpTestPrivilegedDaemonSet`
  - **`ptp-reboot`** — node reboot tests via `ptptesthelper.CreatePtpTestPrivilegedDaemonSet`
- Add `CreatePtpConfigurationsWithRetry` for transient L2/namespace setup failures, wired into both serial and parallel suites.
- Harden `WaitForPtpDaemonToExist` with `Eventually` to tolerate `NotFound` during operator install.
- Add `Eventually` Get for `PtpOperatorConfig` in `EnablePTPEvent`/`PtpEventEnabled` to handle timing during operator bootstrap.
Assisted-by: Cursor

### DIFF
--- a/test/DEVELOPERS.md
+++ b/test/DEVELOPERS.md
@@ -726,7 +726,7 @@ This same process is then used to run a single test.
 The L2 information is populated by calling the following code snippet. Node that this command is destructive and will create new ptpconfigs depending on the desired test mode:
 ```
 // discovers valid ptp configurations based on clock type
-	err := testconfig.CreatePtpConfigurations()
+	err := testconfig.CreatePtpConfigurationsWithRetry(3)
 	if err != nil {
 		Fail(fmt.Sprintf("Could not create a ptp config, err=%s", err))
 	}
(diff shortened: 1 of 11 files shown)


jzding (approved):
/lgtm

PR #227, opened 2026-05-06
OCPBUGS-85046: CI: Skip TGM tests when no WPC NIC found on any cluster node

jzding:
## Summary

Rewrite `PtpConfigTelcoGM` to scan all cluster nodes for WPC-enabled interfaces with GNSS support. If a suitable node is found, use it for TGM tests. If no WPC hardware exists on any node, skip TGM tests instead of failing.

- Scan all nodes deterministically (sorted) for WPC interfaces
- Iterate all interfaces in PHC group to find GNSS, swap to index 0
- Require at least 2 interfaces (`CreatePtpConfigWPCGrandMaster` accesses `ifList[1]`)
- Track transient errors from `GetWPCEnabledInterfaces` and `CheckGNSSForInterface` so the retry loop keeps retrying on pod-not-ready failures instead of skipping prematurely
- Return `ErrNoWPCHardware` sentinel so `BeforeAll` calls `Skip()` instead of `Fail()`
- Remove redundant `BestSolution`/`AlgoTelcoGMString` gate — function is only called for TelcoGrandMasterClock mode

Fixes: [OCPBUGS-85046](https://redhat.atlassian.net/browse/OCPBUGS-85046)

## Test plan

- [ ] Run on cluster with WPC NIC — TGM tests should configure and run normally
- [ ] Run on cluster without WPC NIC — TGM tests should skip with "No WPC NIC found" message
- [ ] Validate `go vet ./test/pkg/testconfig/...` passes

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


jzding:
this is replaced by https://github.com/k8snetworkplumbingwg/ptp-operator/pull/228

PR #229, opened 2026-05-11
feat: GNSS simulation for hardware-free T-GM testing

bnshr:
Adds a GNSS simulator so Telco Grandmaster conformance (`tgm`, `tgmoc`, `tgmbc`) can run in Kind/netdevsim without GNSS hardware.

JIRA: CNF-21632

### Simulator
- `ptp-tools/gnss-sim` — NMEA at 1 Hz, DPLL clock classes 6/7/248, kernel GNSS or PTY output, HTTP API on `:9200` (`/health`, `/api/signal/loss`, `/api/signal/restore`)
- Deployed as a Kubernetes Deployment (`scripts/configGNSS.sh`)
- Design: `docs/netdevsim-ptp-simulation.md`

### Tests and CI
- netdevsim matrix adds `tgm`, `tgmoc`, `tgmbc`; hardware GNSS tests skip when the simulator is active
- Kind CI: crun 1.28, retried kustomize downloads, `DKMS_BRANCH=fix-dashboard`
- DualNIC/HA discovery when Kind strips phc2sys `-r`; DualNICBCHA uses the current phc2sys source; T-GM parallel suites no longer delete the GM mid-run

### Also
- Webhook `GetInterfaces` handles HA profiles with empty `Ptp4lConf`
- `ens3f2` PCI `:13.0` (avoids VRT `:10.0–:12.0` and invalid device 32)
- gomega v1.40.0 → v1.41.0


ptp-tools/Dockerfile.ptpop
```diff
@@ -1,5 +1,5 @@
 FROM docker.io/golang:1.25.7 AS builder
-RUN apt-get update && apt-get install -y binutils-gold && rm -rf /var/lib/apt/lists/*
```
edcdavid, line 2: update is neeeded here

.github/workflows/netdevsim-ci.yml
edcdavid: The branch should be add-gnss to pickup the new gnss + dkms tweaks

scripts/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+    oc login [API_URL]
+
+to start a session against an OpenShift cluster. After login, run `oc` and
+`oc help` to learn more about how to get started with OpenShift.
+
+## License
+
+OpenShift is licensed under the Apache Public License 2.0. The source code for this
+program is [located on github](https://github.com/openshift/oc).
```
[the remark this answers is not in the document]
bnshr: Resolved.

bnshr:
@coderabbitai review

scripts/config.yaml
```diff
@@ -0,0 +1,5 @@
+global:
+  maxoffset: 10000
```
edcdavid: You can remove this file, it is temporary

scripts/run-on-vm.sh
```diff
@@ -317,7 +320,8 @@ if [[ "$RUN_PHASE" == "all" ]]; then
     run_quiet_with_log_dump_on_failure "create-local-registry" ./create-local-registry.sh "$VM_IP"
 
     cd ../ptp-tools
-    run_ptp_tools_parallel_make_step_rows PUSH podman-push push "${_ptp_tool_images[@]}"
+    make podman-cleanall
```
edcdavid: This seems to break the log clean up work done in https://github.com/k8snetworkplumbingwg/ptp-operator/pull/235

edcdavid (commented):
started review. Some initial comments

scripts/run-on-vm.sh
edcdavid: this also leads to images being built twice, with double disk capacity

scripts/run-tests.sh
```diff
@@ -268,12 +268,45 @@ run_ginkgo_suite() {
...
+  else
+    export GNSS_SIM_NMEA_DEVICE="${GNSS_SIM_NMEA_DEVICE:-/var/run/ttyGNSS_TS2PHC}"
+  fi
+  export GNSS_SIM_IFACE1="${GNSS_SIM_IFACE1:-ens1f0}"
+  export GNSS_SIM_IFACE2="${GNSS_SIM_IFACE2:-ens1f1}"
+}
+
+overall_exit=0
 for mode in "${TEST_MODES[@]}"; do
+  init_gnss_sim_env
```
edcdavid: This is the root cause of several existing-mode regressions: `init_gnss_sim_env` runs for **every** `TEST_MODES` entry (`oc`, `bc`, `dualfollower`, `dualnicbcha`, ...), not only `tgm`/`tgmoc`/`tgmbc`.

It always exports `GNSS_SIM_IFACE1`/`GNSS_SIM_NMEA_DEVICE`, so `ptphelper.IsGnssSimConfigured()` is true even when the clock under test is not a simulated T-GM. That then:

- strips phc2sys `-r` via `stripPhc2sysRealtimeInSimulation` (undoes VRT CLOCK_REALTIME)
- skips DualNICBCHA HA and phc2sys SIGTERM with a false "phc2sys is not started" message

Please gate `configGNSS.sh` + these exports on T-GM modes only, and do not set `GNSS_SIM_IFACE1` unless a T-GM scenario is actually running.

scripts/run-on-vm.sh
```diff
@@ -438,6 +454,46 @@ if [[ "$RUN_PHASE" == "all" || "$RUN_PHASE" == "deploy" ]]; then
...
+    GNSS_KERNEL_DEV=""
+    for g in /dev/gnss*; do
+        [ -c "$g" ] && GNSS_KERNEL_DEV="$g" && break
+    done
+    if [ -n "$GNSS_KERNEL_DEV" ]; then
+        export GNSS_SIM_NMEA_DEVICE="${GNSS_SIM_NMEA_DEVICE:-$(basename "$GNSS_KERNEL_DEV")}"
+    else
+        export GNSS_SIM_NMEA_DEVICE="${GNSS_SIM_NMEA_DEVICE:-/var/run/ttyGNSS_TS2PHC}"
+    fi
+    export GNSS_SIM_IFACE1="${GNSS_SIM_IFACE1:-ens1f0}"
```
edcdavid: Same coupling as `run-tests.sh`: gnss-sim is deployed and `GNSS_SIM_IFACE1` is exported **before all modes**, then `run-tests.sh` exports them again per mode.

Existing DualFollower / DualNICBCHA / OC / BC must not inherit GNSS-sim env. Start gnss-sim only when `$TEST_MODES` includes `tgm`/`tgmoc`/`tgmbc` (or pass the env only into those suites).

scripts/run-on-vm.sh
```diff
@@ -294,8 +294,22 @@ run_quiet_with_log_dump_on_failure "install-tools" bash ./install-tools.sh
...
 
+step "Tidying and vendoring Go dependencies"
+run_step_rows_begin "go mod tidy" "go mod vendor"
 
-# ── Images phase (--images) ──────────────────────────────────────────
+run_quiet_with_log_dump_on_failure "go-mod-tidy" go mod tidy
+run_step_row_done "go mod tidy"
+
+run_quiet_with_log_dump_on_failure "go-mod-vendor" go mod vendor
+run_step_row_done "go mod vendor"
```
edcdavid: CI should not `go mod tidy` + `go mod vendor` on every VM run. That mutates the checkout, fights the committed vendor tree, and can hide or create `go.mod` drift (this PR already vendors a gomega bump).

Keep tidy/vendor as a maintainer / PR check (`validate.yml` already runs tidy), not as part of `run-on-vm.sh`.

scripts/run-on-vm.sh
```diff
@@ -438,6 +454,46 @@ if [[ "$RUN_PHASE" == "all" || "$RUN_PHASE" == "deploy" ]]; then
...
     run_ind kubectl get pods -n openshift-ptp -o wide
 
+    SYMLINK_PID=""
+    if [[ "${DKMS_MODE}" == "true" ]]; then
+        bash -c '
+        while true; do
+          for pod in $(kubectl get pods -n openshift-ptp -l app=linuxptp-daemon \
+                       --field-selector=status.phase=Running -o name 2>/dev/null); do
+            kubectl exec -n openshift-ptp ${pod#pod/} -c linuxptp-daemon-container -- \
+              bash -c "for i in 0 1 2 3 4 5 6 7 8 9; do ln -sf nsim_ptp\$i /dev/ptp\$i 2>/dev/null; done" \
```
edcdavid: This background `kubectl exec` loop rewriting `/dev/ptpN` -> `nsim_ptpN` every 5s looks leftover from pre-VRT netdevsim bring-up. VRT / `create-vrt-clocks.sh` already owns the PHC device nodes; racing `ln -sf` from CI can fight the daemon.

Please drop this unless there is a remaining Kind path that still needs it -- and if so, document why `create-vrt-clocks.sh` is insufficient.

test/pkg/ptphelper/ptphelper.go
```diff
@@ -1301,3 +1627,255 @@ func GetLocalClockID(ptpConfig *ptpv1.PtpConfig, profileName string, l2Config l2
...
+	return UseGnssSimulation()
+}
+
+// IsGnssSimConfigured returns true when GNSS simulation env vars are set,
+// indicating the CI environment has a gnss-sim instance available.
+func IsGnssSimConfigured() bool {
+	_, hasDevice := os.LookupEnv("GNSS_SIM_NMEA_DEVICE")
+	_, hasIface := os.LookupEnv("GNSS_SIM_IFACE1")
+	return hasDevice || hasIface
+}
```
edcdavid: `LookupEnv("GNSS_SIM_IFACE1")` is true whenever the var is **set**, including the default `ens1f0` that `init_gnss_sim_env` / `run-on-vm.sh` export for every mode.

This is not a reliable "gnss-sim is the clock under test" signal. After gating GNSS setup to T-GM modes, prefer `UseGnssSimulation()` (actual `/dev/gnssN` or PTY discovery) for test skips, and do not treat `GNSS_SIM_IFACE1` as sufficient.

test/pkg/ptphelper/ptphelper.go
```diff
@@ -1301,3 +1627,255 @@ func GetLocalClockID(ptpConfig *ptpv1.PtpConfig, profileName string, l2Config l2
...
+// UseGnssSimulation is true when a kernel virtual GNSS device or gnss-sim PTY is in use
+// (i.e. not a physical E810 with real external GNSS hardware).
+func UseGnssSimulation() bool {
+	return gnssSimNmeaActive.Load()
+}
+
+// IsGnssSimulatedCI is true for conformance paths that target a virtual/simulated GNSS source.
+func IsGnssSimulatedCI() bool {
+	return UseGnssSimulation()
+}
```
edcdavid: Redundant alias of `UseGnssSimulation()`. Callers in `ptp.go` can use that directly. Three overlapping predicates (`UseGnssSimulation`, `IsGnssSimulatedCI`, `IsGnssSimConfigured`) will drift -- please collapse to one that reflects actual GNSS-sim discovery, not env leakage.

test/conformance/serial/ptp.go
```diff
@@ -1064,6 +1081,9 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 			})
 
 			It("DualNICBCHA phc2sys switches to secondary ptp4l when primary interface fails", func() {
+				if ptphelper.IsGnssSimConfigured() {
+					Skip("phc2sys is not started in netdevsim/Kind simulation (shared host CLOCK_REALTIME)")
+				}
```
edcdavid: This skip is incorrect, and it fires for every netdevsim job once `GNSS_SIM_IFACE1` is exported.

phc2sys **is** started; `stripPhc2sysRealtimeInSimulation` only removes `-r`. DualNICBCHA HA on main/VRT does not need this skip (see passing DualNICBCHA on current main).

Please delete the skip (or gate it on T-GM-only if there is a real reason). Leaving it in means DualNICBCHA is never exercised in this CI.

test/conformance/serial/ptp.go
```diff
@@ -2409,6 +2474,9 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 			})
 
 			It("Should restart phc2sys with no socket errors after SIGTERM", func() {
+				if ptphelper.IsGnssSimConfigured() {
+					Skip("phc2sys is not started in netdevsim/Kind simulation (shared host CLOCK_REALTIME)")
+				}
```
edcdavid: Same false premise as the DualNICBCHA skip: phc2sys is still launched; only `-r` is stripped. This skips a useful SIGTERM check for every mode in GNSS CI, including OC/BC.

Drop the skip, or restrict it to simulated T-GM if phc2sys is truly omitted there.

test/conformance/serial/ptp.go
```diff
@@ -3680,21 +4534,25 @@ func checkProcessStatus(fullConfig testconfig.TestConfig, state string) {
...
+	}
 
 	time.Sleep(10 * time.Second)
 	buf, _, _ := pods.ExecCommand(client.Client, true, fullConfig.DiscoveredClockUnderTestPod, pkg.PtpContainerName, []string{"curl", pkg.MetricsEndPoint})
 	ret, err := processRunning(buf.String(), state)
 	Expect(err).To(BeNil())
-	Expect(ret["phc2sys"]).To(BeTrue(), fmt.Sprintf("Expected phc2sys to be  %s for GM", state))
+	if !ptphelper.IsGnssSimConfigured() {
+		Expect(ret["phc2sys"]).To(BeTrue(), fmt.Sprintf("Expected phc2sys to be  %s for GM", state))
+	}
```
edcdavid: These `IsGnssSimConfigured()` guards skip phc2sys process-status checks for **all** modes in this CI, while phc2sys is still running without `-r`.

If the intent is "do not require gpsd/gpspipe on simulated T-GM", gate on `UseGnssSimulation()` **and** T-GM discovery -- not on GNSS env vars that are always set.

test/conformance/serial/ptp.go
```diff
@@ -4194,6 +5072,88 @@ func checkClockClassStateReturnBool(fullConfig testconfig.TestConfig, expectedSt
...
+// can target any pod (GM, BC, OC) independently.
+func checkClockClassInMetrics(metricsOutput string, expectedState string) bool {
+	scanner := bufio.NewScanner(strings.NewReader(metricsOutput))
+	for scanner.Scan() {
+		line := scanner.Text()
+		if matches := clockClassRe.FindStringSubmatch(line); matches != nil {
+			process := matches[2]
+			class := matches[3]
+			if strings.TrimSpace(process) == "ptp4l" && strings.TrimSpace(class) == expectedState {
+				return true
```
edcdavid: This only accepts `process="ptp4l"`. On T-GM the clock class that tracks GNSS lock is often published on **ts2phc**, while ptp4l may stay at a different class.

`waitForWPCGMReady()` above depends on this helper. Please also match `ts2phc` (or any process reporting the expected class) or T-GM wait/holdover checks will flake or pass for the wrong clock.

test/pkg/event/event.go
```diff
@@ -764,8 +764,13 @@ func createStoredEvent(data []byte) (aStoredEvent exports.StoredEvent, aType str
...
 	for _, v := range d.Values {
-		key := compositeEventKey(v.Resource, string(v.DataType))
-		values[key] = v.Value
+		dt := string(v.DataType)
+		// Unique resource/data_type key (upstream event keying).
+		values[compositeEventKey(v.Resource, dt)] = v.Value
+		// Also index by data_type so helpers can find "metric"/"notification".
+		if dt != "" {
+			values[dt] = v.Value
+		}
```
edcdavid: Last-wins `values[dt]` keys (`"metric"` / `"notification"`) are not needed: `verifyEvent` already walks all `EventValues`. If two values share a data type, this overwrites the first.

Please drop the extra index unless there is a caller that cannot use `compositeEventKey`.

docs/netdevsim-ptp-simulation.md
```diff
@@ -0,0 +1,595 @@
...
+        gm1 -->|"PTP over VLAN 1500"| oc1
+    end
+```
+
+```mermaid
+flowchart LR
+    subgraph tgmbc_topo ["TGMBC Topology"]
+        gm2["WPC T-GM<br/>ens1f0 (worker)"]
+        bc_slave["BC Slave<br/>ens3f0 (worker2)"]
+        bc_master["BC Master<br/>ens3f1 (worker2)"]
```
edcdavid: TGMBC master is `ens3f2` on VLAN 1502 (same PHC as ens3f0/ens3f1, PCI `:13.0`). The diagram still shows `ens3f1` as BC master, which is the DualFollower pair on VLAN 1500.

Please update this (and any VLAN labels) so the extra port purpose is obvious.

scripts/gnss-sim-pod.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: v1
```
edcdavid: Dead manifest: CI applies `scripts/gnss-sim-deployment.yaml` via `configGNSS.sh`. This Pod YAML is never used. Please delete it (or switch the deploy path to one object) so the two specs cannot drift.

test/pkg/consts.go
```diff
@@ -23,6 +23,7 @@ const (
 	// policy name
 	PtpGrandMasterPolicyName    = "test-grandmaster"
 	PtpWPCGrandMasterPolicyName = "test-wpc-grandmaster"
+	PtpSimGrandMasterPolicyName = "test-sim-grandmaster"
```
edcdavid: `test-sim-grandmaster` is never created. Simulated T-GM uses `PtpWPCGrandMasterPolicyName` (`test-wpc-grandmaster`). This constant (and the matching `clean.Configs()` entry) looks leftover -- please drop unless a follow-up actually creates that policy.

test/pkg/ptptesthelper/ptptesthelper.go
```diff
@@ -441,9 +441,10 @@ func (p *PortEngine) TurnPortUp(port string) error {
 		[]string{"ip", "link", "set", port, "up"})
 
 	logrus.Infof("Turning interface: %s in pod %s up, stdout: %s, stderr: %s", port, p.ClockPod.Name, stdout.String(), stderr.String())
-	if err == nil {
-		p.nmSetManaged(port, true)
-	}
+	// NM is NOT re-enabled here: the DeferCleanup registered in TurnPortDown
+	// handles that after the test completes. Re-enabling NM eagerly can cause
+	// it to re-activate the interface (DHCP, link detection) and interfere
+	// with ptp4l's port state machine, preventing SLAVE recovery.
```
edcdavid: Unrelated to GNSS simulation, and the DualFollower holdover / BMCA port-down changes that motivated it were reverted.

Please keep NM re-enable here (or split this into its own PR with DualFollower evidence). As written it changes port-recovery behavior for every mode.

scripts/ptpswitchconfig.cfg
```diff
@@ -6,6 +6,7 @@
 [ens2f1]
 [ens3f0]
 [ens3f1]
+[ens3f2]
```
edcdavid: `ens3f2` is added here, but this file is no longer installed (switch ptp4l BC was replaced by transparent OVS). The extra port belongs in Kind/netdevsim + VLAN 1502 setup, not in an unused switch ptp4l config.

Please drop the edit or restore a caller so this file is not a trap.

api/v1/ptpconfig_webhook.go
```diff
@@ -460,9 +460,17 @@ func GetInterfaces(config PtpConfig, mode PtpRole) (interfaces []string) {
...
 		return interfaces
 	}
+	// HA-only profiles (phc2sys + haProfiles) intentionally omit Ptp4lConf.
+	ptp4lConf := config.Spec.Profile[0].Ptp4lConf
+	if ptp4lConf == nil || strings.TrimSpace(*ptp4lConf) == "" {
+		if mode == Slave && config.Spec.Profile[0].Interface != nil {
+			return []string{*config.Spec.Profile[0].Interface}
+		}
+		return interfaces
+	}
```
edcdavid: Optional HA robustness, not required for GNSS simulation. `GetInterfaces` on an HA-only profile is a pre-existing edge case.

If DualNICBCHA already passes on main/VRT without this, consider moving it (and `ptpconfig_webhook_test.go`) to a separate PR so this GNSS change stays reviewable.

test/pkg/testconfig/testconfig.go
```diff
@@ -971,22 +1055,56 @@ func CreatePtpConfigGrandMaster(nodeName, ifName string) error {
...
+// would make every DualNIC BC look secondary (IsSecondaryBc) and break
+// DualNICBC / DualNICBCHA discovery — keep a non-nil opts string instead.
+// Baremetal keeps -r so phc2sys still syncs the per-node system clock.
+func stripPhc2sysRealtimeInSimulation(phc2sysOpts *string) *string {
+	if phc2sysOpts == nil || !ptphelper.IsGnssSimConfigured() {
+		return phc2sysOpts
+	}
+	stripped := stripPhc2sysRealtimeOpts(*phc2sysOpts)
+	logrus.Infof("Stripping phc2sys -r in netdevsim/Kind simulation (shared host CLOCK_REALTIME); opts=%q", stripped)
+	return &stripped
```
edcdavid: This was true before VRT: Kind workers shared the host CLOCK_REALTIME, so `phc2sys -r` fought ts2phc/gnss-sim.

Each worker now has its own CLOCK_REALTIME, so we should keep `-r` in every mode — including simulated T-GM. Stripping it here is also why DualNICBCHA HA, phc2sys SIGTERM, and T-GM CLOCK_REALTIME checks get skipped further down.

If a no-RT path is still needed, that's `DisableAllSlaveRTUpdate` (non-VRT Kind only), not `IsGnssSimConfigured()`.

test/pkg/testconfig/testconfig.go
```diff
@@ -564,6 +588,10 @@ func createPtpConfigurations(ctx context.Context) error {
 	logrus.Tracef("L2DiscoveryConfig: %s\n", config)
 	logrus.Tracef("L2 ifListFiltered=%+v, ifListUnfiltered=%+v", config.GetPtpIfList(), config.GetPtpIfListUnfiltered())
 	GlobalConfig.L2Config = config
+	ptphelper.NormalizeL2IntegratedGnssNICsForTelcoGM()
+	if ptphelper.IsGnssSimConfigured() && !ptphelper.L2ConfigReportsIntelWPC(config) {
+		ptphelper.ApplyIntegratedGnssSimWPCPCIOverlay()
+	}
```
edcdavid, line 610: This overlay should only run for T-GM (`tgm` / `tgmoc` / `tgmbc`).

Right now it fires whenever gnss-sim is present, so DualNICBC gets solved onto worker1's GNSS NICs (`ens1f*`) and CLOCK_REALTIME never locks on the BC. Please gate it on the desired PTP mode.

test/conformance/serial/ptp.go
edcdavid: Please drop this skip. Once we keep `phc2sys -r`, HA phc2sys does start in Kind. Skipping the test just hides whether failover actually works.

test/conformance/serial/ptp.go
```diff
@@ -1091,31 +1121,58 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
...
+						if len(primaryBCSlaveInterfaces) == 1 {
+							selectedInterface = primaryBCSlaveInterfaces[0]
+							logMatches = nil
+							logrus.Warnf("no phc2sys selecting line in SinceTime/TailLines; assuming HA source %s (sole primary BC slave)", selectedInterface)
+							return nil
+						}
+						if err != nil {
+							return err
+						}
+						return fmt.Errorf("no phc2sys selecting line yet (primary slaves=%v)", primaryBCSlaveInterfaces)
```
edcdavid: I'd rather fail than guess the HA source when the log line isn't there.

DualNICBCHA already works on main with `GetPodLogsRegex` (full log, last match). The SinceTime / TailLines / "assume the only slave" path is a Kind logging workaround, not GNSS, and it can take the wrong interface down.

test/conformance/serial/ptp.go
edcdavid: Please don't skip this. phc2sys is running — including DualFollower and simulated T-GM — once we keep `-r`.

The DualFollower failure we actually hit (`openshift_ptp_process_status{process="phc2sys"}==0` for minutes, while CLOCK_REALTIME was already moving) is a linuxptp-daemon bug, not GNSS:

- [#215](https://github.com/k8snetworkplumbingwg/linuxptp-daemon/pull/215) waits to start phc2sys until `|ptp4l offset| < 1s`
- `/ready` still treats that delayed process as missing, so `/emit-logs` can run first
- status is then published under the on-disk name `phc2sys.N.config=0`, while live logs use `ptp4l.N.config` — Prometheus keeps the stale 0

That's [linuxptp-daemon#269](https://github.com/k8snetworkplumbingwg/linuxptp-daemon/pull/269). Keep the test; it should pass once that daemon change is in.

test/conformance/serial/ptp.go
```diff
@@ -2977,6 +3065,575 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
...
+						return ptphelper.GNSSSimIsHealthy()
+					}, pkg.TimeoutIn1Minute, 5*time.Second).Should(BeTrue(),
+						"gnss-sim health check failed")
+				})
+
+				gmPod := getGMPod()
+
+				// phc2sys -r is stripped in Kind/netdevsim (shared host CLOCK_REALTIME).
+				By("checking sim GM required processes status (ts2phc, ptp4l)", func() {
+					processesArr := [...]string{"ts2phc", "ptp4l"}
```
edcdavid: Simulated T-GM still runs phc2sys (with `-r`). Please add it here and in the process metrics.

gpsd/gpspipe can stay skipped — gnss-sim owns `/dev/gnssN`. Also worth waiting for CLOCK_REALTIME to reach LOCKED.

test/conformance/serial/ptp.go
```diff
@@ -3680,21 +4534,25 @@ func checkProcessStatus(fullConfig testconfig.TestConfig, state string) {
...
+			"phc2sys process status not detected")
+	}
 
 	time.Sleep(10 * time.Second)
 	buf, _, _ := pods.ExecCommand(client.Client, true, fullConfig.DiscoveredClockUnderTestPod, pkg.PtpContainerName, []string{"curl", pkg.MetricsEndPoint})
 	ret, err := processRunning(buf.String(), state)
 	Expect(err).To(BeNil())
-	Expect(ret["phc2sys"]).To(BeTrue(), fmt.Sprintf("Expected phc2sys to be  %s for GM", state))
+	if !ptphelper.IsGnssSimConfigured() {
+		Expect(ret["phc2sys"]).To(BeTrue(), fmt.Sprintf("Expected phc2sys to be  %s for GM", state))
```
edcdavid: Same as the SIGTERM skip: please don't special-case GNSS sim here. T-GM should show `phc2sys` process_status 1.

If DualFollower still flaps at 0 with CLOCK_REALTIME already moving, that's the daemon emit-logs label bug ([linuxptp-daemon#269](https://github.com/k8snetworkplumbingwg/linuxptp-daemon/pull/269)), not a reason to skip the check.

test/conformance/serial/ptp.go
```diff
@@ -1465,7 +1527,9 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 				ptptesthelper.VerifyClockClassViaPMC(fullConfig, "/var/run/ptp4l.0.config", int(fbprotocol.ClockClass6))
 
 				By(fmt.Sprintf("Verifying initial clockClass is %s via Event API", expectedClockClassStr))
-				verifyClockClassCurrentState(int(expectedClockClass), 60*time.Second)
+				// Allow extra time on TGMOC where the consumer runs on the OC node
+				// and initial clock-class push can lag behind metrics.
+				verifyClockClassCurrentState(int(expectedClockClass), 120*time.Second)
```
edcdavid: This flakes on TGMBC. After cloud-event-proxy restarts, the first `getCurrentState` is often 255 or 248 even though ptp4l metrics already show class 6. A flake retry then starts with the BC still unlocked.

Please wait for the GM to be ready first (`waitForWPCGMReady`), and poll `getCurrentState` until the class matches instead of taking the first event.

test/conformance/parallel/parallel_suite_test.go
```diff
@@ -54,18 +53,42 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
+	// serial on the same cluster that wipe deletes PtpConfigs mid-serial and
+	// cascades into nil-pod panics (seen on tgm-serial vs tgm-parallel).
+	desired := testconfig.GetDesiredConfig(false)
+	existing := testconfig.GetFullDiscoveredConfig(pkg.PtpLinuxDaemonNamespace, true)
+	reusedSerialConfigs := existing.Status == testconfig.DiscoverySuccessStatus &&
+		existing.PtpModeDiscovered == desired.PtpModeDesired &&
+		existing.DiscoveredClockUnderTestPod != nil
+	if reusedSerialConfigs {
+		logrus.Infof("Reusing existing PTP configs for mode %s (skipping clean/create)", desired.PtpModeDesired)
+		fullConfig = existing
```
edcdavid: I don't think serial and parallel actually overlap, so we shouldn't skip create/restart to "reuse" serial's configs.

Ginkgo runs the two packages one after the other. Kind CI only runs serial. The OpenShift release job (`ginkgo -p serial parallel` on the telco5g baremetal cluster) also finishes serial AfterSuite before parallel starts.

On that job `KEEP_PTPCONFIG` is unset, so serial already wiped the PtpConfigs. If parallel skips create, T-GM parallel has no GM. Please always create, discover, and restart here.

test/conformance/parallel/parallel_suite_test.go
```diff
@@ -115,7 +138,11 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 		"clock-under-test pod missing; label node with "+pkg.PtpClockUnderTestNodeLabel)
 })
 var _ = AfterSuite(func() {
+	// Do not clean.All() here. Parallel often overlaps the serial suite for the
+	// same PTP_TEST_MODE on one cluster; wiping configs from AfterSuite deletes
+	// serial's PtpConfigs mid-run. Serial AfterSuite / next mode's create path
+	// owns cleanup.
 	if DeletePtpConfig {
-		clean.All()
+		logrus.Info("parallel AfterSuite: skipping clean.All() to avoid racing serial suite")
```
edcdavid: Same overlap assumption as above — parallel AfterSuite does not run while serial is still going. Please `clean.All()` when `DeletePtpConfig` is true.

(It's also never initialized in this suite, so this skip doesn't do anything today.)

test/pkg/pods/pods.go
```diff
@@ -126,134 +125,108 @@ func WaitForCondition(cs *testclient.ClientSet, pod *corev1.Pod, conditionType c
...
+// snapshotPodLogsRegex reads a non-follow log snapshot and returns every regex
+// match. Callers that need the current value must use matches[len(matches)-1].
+//
+// Non-follow only: Follow + find-first returns the oldest hit in the stream,
+// which is wrong for "current phc2sys HA source" when daemon logs are large.
+// TailLines (or SinceTime) bounds the read so the client can finish the stream.
+//
+// Incomplete reads are hard errors: the API delivers oldest→newest, so a
+// truncated buffer's last match is not the latest line and must not be trusted.
+func snapshotPodLogsRegex(namespace, podName, containerName string, r *regexp.Regexp, tailLines *int64, since *time.Time) (matches [][]string, err error) {
```
edcdavid: Can we keep main's `GetPodLogsRegex`? DualNICBCHA already finds the latest `phc2sys selecting` line from a full log read. The TailLines / SinceTime snapshot (and `log_regex_test.go`) aren't needed for GNSS.

scripts/run-on-vm.sh
```diff
@@ -438,6 +454,46 @@ if [[ "$RUN_PHASE" == "all" || "$RUN_PHASE" == "deploy" ]]; then
...
+        ' &
+        SYMLINK_PID=$!
+        echo "Symlink maintainer PID: $SYMLINK_PID"
+        cleanup_symlink() { [[ -n "$SYMLINK_PID" ]] && kill "$SYMLINK_PID" 2>/dev/null || true; }
+        trap cleanup_symlink EXIT
+    fi
+
+    # Start GNSS simulator as a pod for T-GM simulation tests
+    export GNSS_SIM_IMAGE="${IMG_PREFIX}:gnss-sim"
+    ./configGNSS.sh
```
edcdavid: `run-tests.sh` already starts gnss-sim. Calling `configGNSS.sh` here deploys it twice.

Please leave deploy to `run-tests.sh`, and do it once for the whole Kind cluster so any mode can apply a T-GM config without changing topology.

scripts/run-tests.sh
edcdavid: Can we call this once before the mode loop? Redeploying between oc / bc / tgm isn't needed and can bounce the pod as the next suite starts.

scripts/run-on-vm.sh
```diff
@@ -294,8 +294,22 @@ run_quiet_with_log_dump_on_failure "install-tools" bash ./install-tools.sh
...
 PS1="${PS1:-}" source ~/.bashrc
 
+step "Tidying and vendoring Go dependencies"
+run_step_rows_begin "go mod tidy" "go mod vendor"
 
-# ── Images phase (--images) ──────────────────────────────────────────
+run_quiet_with_log_dump_on_failure "go-mod-tidy" go mod tidy
+run_step_row_done "go mod tidy"
+
+run_quiet_with_log_dump_on_failure "go-mod-vendor" go mod vendor
```
edcdavid: This isn't GNSS — main is already vendored, and running tidy in CI has bitten us. Please drop it.

scripts/run-on-vm.sh
```diff
@@ -438,6 +454,46 @@ if [[ "$RUN_PHASE" == "all" || "$RUN_PHASE" == "deploy" ]]; then
...
+        while true; do
+          for pod in $(kubectl get pods -n openshift-ptp -l app=linuxptp-daemon \
+                       --field-selector=status.phase=Running -o name 2>/dev/null); do
+            kubectl exec -n openshift-ptp ${pod#pod/} -c linuxptp-daemon-container -- \
+              bash -c "for i in 0 1 2 3 4 5 6 7 8 9; do ln -sf nsim_ptp\$i /dev/ptp\$i 2>/dev/null; done" \
+              2>/dev/null || true
+          done
+          sleep 5
+        done
+        ' &
```
edcdavid: This `/dev/ptpN` → `nsim_ptpN` loop isn't needed for GNSS (or for netdevsim on main). Please remove it.

scripts/run-on-vm.sh
```diff
@@ -335,7 +349,8 @@ if [[ "$RUN_PHASE" == "all" ]]; then
 
     step "Building kustomize"
     cd ..
-    run_quiet_with_log_dump_on_failure "make-kustomize" make kustomize
+    # GitHub release downloads can flake under parallel matrix jobs; retry make.
+    run_quiet_with_log_dump_on_failure "make-kustomize" bash -c 'for i in 1 2 3; do make kustomize && exit 0; echo "make kustomize attempt $i failed"; sleep $((i*3)); done; exit 1'
```
edcdavid: The Makefile already retries the kustomize download. `make kustomize` here is enough.

.github/workflows/netdevsim-ci.yml
```diff
@@ -162,26 +169,39 @@ jobs:
...
 
       - name: Copy checkout to /root/ptp-operator
         run: sudo cp -a "$GITHUB_WORKSPACE" /root/ptp-operator
 
       - name: "Build images and export tarball"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          sudo bash -l <<'BUILD'
+          sudo GITHUB_TOKEN="$GITHUB_TOKEN" bash -l <<'BUILD'
```
edcdavid: We don't need to pass `GITHUB_TOKEN` through sudo — main doesn't, and curl isn't sending it for these downloads. Same for the LOAD / DEPLOY heredocs below.

ptp-tools/Makefile
```diff
@@ -51,9 +51,20 @@ podman-push-%:
...
 else
 podman-push-%:
-	podman manifest push ${IMG_PREFIX}:$* docker://${IMG_PREFIX}:$*
+	podman push ${IMG_PREFIX}:$* docker://${IMG_PREFIX}:$*
 endif
 
+# Build then immediately push each image sequentially.
+# Using -t (plain image tag) instead of --manifest avoids podman 5.x bugs
+# where manifest-list entries become unresolvable in local storage.
+podman-buildpushall: $(foreach v, $(VALUES), podman-buildpush-$(v))
```
edcdavid: `podman-buildpushall` isn't GNSS. Please keep main's manifest build/push (and just add `gnss-sim` to VALUES).

scripts/create-local-registry.sh
```diff
@@ -11,14 +11,26 @@ openssl req -x509 -newkey rsa:4096 -sha256 -days 3650 -nodes  -keyout ~/registry
...
 
+# Create the containerd certificate trust bundle that Kind nodes will mount at
+# /etc/containerd/certs.d/$VM_IP/ (via kind-config.yaml extraMounts).
+cp ~/registry/registry.crt ~/registry/ca.crt
+cat > ~/registry/hosts.toml <<EOF
+server = "https://$VM_IP"
+
+[host."https://$VM_IP"]
+  ca = "/etc/containerd/certs.d/$VM_IP/ca.crt"
+EOF
```
edcdavid: Kind already mounts `~/registry` into containerd certs.d. This `hosts.toml` / CA rewrite isn't GNSS — please revert.

scripts/gnss-sim-pod.yaml
```diff
@@ -0,0 +1,72 @@
...
+kind: Pod
+metadata:
+  name: gnss-sim
```
edcdavid: Nothing uses this file — `configGNSS.sh` applies `gnss-sim-deployment.yaml`. Safe to delete.

test/pkg/event/event.go
edcdavid: Main only keys events by `resource/dataType` (PR 281). Indexing by bare `data_type` as well can collide when two resources both have `metric` / `notification`. Not GNSS — please drop the extra map entry.

test/pkg/consts.go
edcdavid: Nothing creates `test-sim-grandmaster`; simulated T-GM uses the normal WPC GM policy. Please remove this constant and the matching `clean.All` entry.

PR #230, opened 2026-05-11
Set phc2sysOpts to nil when all slave RT clock updates are disabled 

edcdavid:
Improving disabling phc2sys updates for netdevsim simulation by passing a nil configuration instead of "-v" to print version and quit. Improves reliability of test: "PTP ClockSync Can provide a profile with higher priority"

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated configuration handling so related timing settings no longer force a hardcoded fallback when a specific update mode is disabled.
  * Improved support for high-availability setups by adjusting command options more safely and consistently.
  * Preserved existing option values while removing an incompatible flag, reducing unexpected behavior in non-grandmaster node configurations.


jzding (dismissed):
/lgtm

PR #223, opened 2026-04-28
test: Enable PTP authentication tests in CI with --auth flag

gtannous-spec:
## Summary                                                                   
                                                                                                                                                                                  
 - Add `--auth <true|false>` flag to `run-tests.sh` that applies PTP security                                                                                                    
    config and enables switch authentication before test execution                                                                                                                
 - Add SPP mismatch negative test that verifies `auth: bad message` logs when                                                                                                    
    GM and slave have mismatched SPP values and that slave doesn't enter LOCKED state anymore.                                                                                                                                    
 - Enable auth tests by default in the VM CI pipeline (`run-on-vm.sh`)                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
## Test plan                                                                                                                                                                    
                                                                                                                                                                                  
 - [x] Run `./run-tests.sh --kind serial --mode oc --auth true` with auth-capable                                                                                                
        hardware and verify the SPP mismatch test passes                       
 - [x] Run `./run-tests.sh --kind serial --mode oc` (no `--auth`) and verify auth                                                                                                
        tests are skipped                                                                                                                                                         
 - [x] Confirm VM CI pipeline (`run-on-vm.sh`) completes with auth tests enabled

PR #231, opened 2026-05-13
Fix phc2sys source failover test to handle multiple log matches

gtannous-spec:
## Summary

The phc2sys source failover test assumed exactly one log match for the
source interface pattern, causing false failures when phc2sys logs
multiple source selection lines (e.g., during restarts or re-syncs).

- Changed assertion from `== 1` to `>= 1` to accept multiple matches
- Use the last match instead of the first, which reflects the most
  recent source interface selection
- Added logging for selected interface names to aid debugging

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -1031,10 +1031,12 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 					fullConfig.DiscoveredClockUnderTestPod.Name, pkg.PtpContainerName,
 					phc2sysLogPattern, false, pkg.TimeoutIn1Minute)
 				Expect(err).NotTo(HaveOccurred())
-				Expect(len(logMatches)).To(BeNumerically("==", 1), "Could not identify which interface phc2sys is using")
+				Expect(len(logMatches)).To(BeNumerically(">=", 1), "Could not identify which interface phc2sys is using")
 
-				logrus.Infof("phc2sys log matching line: %v", logMatches[0][0])
-				selectedInterface = logMatches[0][1]
+				lastMatch := logMatches[len(logMatches)-1]
+				logrus.Infof("phc2sys log matching line: %v", lastMatch[0])
+				selectedInterface = lastMatch[1]
+				logrus.Infof("selected interface: %s", selectedInterface)
 
 				// Save it as primary interface
 				primaryInterface := selectedInterface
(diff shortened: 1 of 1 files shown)


test/conformance/serial/ptp.go
```diff
@@ -1031,10 +1031,12 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
...
 				Expect(err).NotTo(HaveOccurred())
-				Expect(len(logMatches)).To(BeNumerically("==", 1), "Could not identify which interface phc2sys is using")
+				Expect(len(logMatches)).To(BeNumerically(">=", 1), "Could not identify which interface phc2sys is using")
 
-				logrus.Infof("phc2sys log matching line: %v", logMatches[0][0])
-				selectedInterface = logMatches[0][1]
+				lastMatch := logMatches[len(logMatches)-1]
+				logrus.Infof("phc2sys log matching line: %v", lastMatch[0])
+				selectedInterface = lastMatch[1]
+				logrus.Infof("selected interface: %s", selectedInterface)
```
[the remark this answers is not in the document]
gtannous-spec: Optional: Will modify that after we validate all ci-tests pass.

test/conformance/serial/ptp.go
```diff
@@ -1031,10 +1031,12 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 					fullConfig.DiscoveredClockUnderTestPod.Name, pkg.PtpContainerName,
 					phc2sysLogPattern, false, pkg.TimeoutIn1Minute)
 				Expect(err).NotTo(HaveOccurred())
-				Expect(len(logMatches)).To(BeNumerically("==", 1), "Could not identify which interface phc2sys is using")
+				Expect(len(logMatches)).To(BeNumerically(">=", 1), "Could not identify which interface phc2sys is using")
```
nocturnalastro, line 1034: Why should we see more than 1? 

Last time it was more than 1 we had a bug that was duplicating lines. 

The other reason would be it is switching master then switching again which would also be odd. 

Do you have an example of this happening?

gtannous-spec: This test is causing this https://github.com/k8snetworkplumbingwg/ptp-operator/pull/223 PR to fail on this test (after enabling PTP authentication), I added there a CI negative test to verify that slaves don't sync to master in case we have keys mismatch. 

while we change SPP from 1 to 0 on the GM config and then from 0 to 1 each change causes the daemon to restart the ptp4l/phc2sys processes inside the same pod.
phc2sys re-selects its interface each time and logs "selecting". By the time the DualNICBCHA test `("DualNICBCHA phc2sys switches to secondary ptp4l when primary interface fails")` runs, the pod has 3 accumulated entries: 1 from initial startup + 1 from spp change + 1 from spp restore.

I was not tracking the bug you talked about following the fact that "was duplicating lines", if this could lead to another issue then I can try a different approach to "get rid" of the 3 log entries.

gtannous-spec: I fixed that PR, now all tests pass. I guess problem was replacing the pod before it was in LOCKED state. 
which leads the operator to reconcile causing another process to restart so pod ends up selecting multiple entries. 

Will close this one.

gtannous-spec:
Flaky Ci-test failures were fixed for [PR#223](https://github.com/k8snetworkplumbingwg/ptp-operator/pull/223) , 
will close this PR !

PR #235, opened 2026-05-19
CNF-22218: Clean and organize test CI output

shirmoran:
- Add stepped progress with multi-row blocks. When /dev/tty is available print pending step labels once and rewrite the block in place to append OK as each step finishes.
-  Parallel ptp-tools image build/push/clean.
- Add timestamps logging.
- Runs commands quietly and dump acptured logs only on failure.
- Add a bash filter that wraps each Ginkgo spec with START and DONE + STATUS (with coloring) lines.
- Enable SilenceSkips in serial and parallel suite reporters so Ginkgo omits redundant end-of-spec skip blocks. In addition, suppress the verbose EmitFailure skip line.

### DIFF
--- a/scripts/run-on-vm.sh
+++ b/scripts/run-on-vm.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -x
 set -euo pipefail
 
 export DKMS_MODE="${DKMS_MODE:-false}"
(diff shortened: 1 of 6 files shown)


scripts/run-on-vm.sh
```diff
@@ -109,30 +366,46 @@ if [[ "$RUN_PHASE" == "all" || "$RUN_PHASE" == "deploy" ]]; then
...
+    run_quiet_with_log_dump_on_failure "k8s-start" ./k8s-start.sh "$VM_IP"
+    run_step_row_done "Building kind cluster..."
+    run_step_rows_end
 
-    cd ../ptp-tools
-    make deploy-all || true
-    sleep 5
+    step "Deploying ptp-operator manifests"
+    cd "${PTP_TOOLS_DIR}"
+    run_quiet_with_log_dump_on_failure "make-deploy-all" make deploy-all || true
```
edcdavid: The first time it is expected to fail. "|| true" should be in the command :  
run_quiet_with_log_dump_on_failure "make-deploy-all" sh -c "make deploy-all || true"

edcdavid (approved):

PR #240, opened 2026-05-27
ci: fix podman manifest push "image not known" errors

edcdavid:
## Summary
- Remove `podman image rm` from the `clean-image` Makefile target — only the manifest needs clearing before rebuild. Removing images triggers storage GC that invalidates manifest references for other images built in parallel.
- Add `concurrency` guard to cancel previous CI runs when a new push arrives on the same PR.

## Test plan
- [ ] Verify netdevsim CI `ptp-images` job passes without "image not known" errors
- [ ] Verify pushing to a PR cancels any in-progress CI run

### DIFF
--- a/.github/workflows/netdevsim-ci.yml
+++ b/.github/workflows/netdevsim-ci.yml
@@ -8,6 +8,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+concurrency:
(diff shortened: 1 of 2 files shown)


jzding (approved):
/lgtm

PR #241, opened 2026-05-28
ci: improve netdevsim CI robustness and portability

edcdavid:
## Summary

- Detect CA cert directory by path instead of relying on DKMS_MODE variable
- Detect OS family from /etc/os-release for package manager selection (apt vs dnf)
- Remove symlink maintainer loop (no longer needed after symlink removal commit)

## Test plan

- [ ] Run netdevsim CI on Ubuntu (DKMS) VM
- [ ] Run netdevsim CI on Fedora/RHEL VM

### DIFF
--- a/scripts/install-tools.sh
+++ b/scripts/install-tools.sh
@@ -12,15 +12,27 @@ else
     exit 1
 fi
 
+# Detect OS family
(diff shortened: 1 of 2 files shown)


nocturnalastro (approved):
/lgtm

PR #246, opened 2026-06-03
ci: add podman-build-and-pushall target to ptp-tools

edcdavid:
## Summary
- Add per-image build targets to `ptp-tools/Makefile` that sequentially clean, build, and save/push each image one by one
- New targets: `podman-build-and-saveall` (clean→build→save per image), `podman-build-and-pushall` (clean→build→push per image), `podman-saveall` / `podman-save-%`
- Each image is saved to its own file under `/tmp/ptp-images/<tag>.tar` (configurable via `SAVE_DIR`)
- `--images` phase no longer creates a registry or pushes — it only builds and saves
- `--load` now loads from per-image tar files in the given directory
- Default `all` phase uses `podman-build-and-pushall` for the full build→push→deploy flow
- Add missing `.PHONY` declarations

## Test plan
- [ ] `make podman-build-and-saveall` — verify each image is built and saved to `/tmp/ptp-images/<tag>.tar`
- [ ] `run-on-vm.sh --images <VM_IP>` — verify images are built and saved without registry/push
- [ ] `run-on-vm.sh --load /tmp/ptp-images <VM_IP>` — verify per-image tarballs are loaded, retagged, and pushed
- [ ] `run-on-vm.sh <VM_IP>` — verify default flow builds, pushes, and deploys end-to-end

### DIFF
--- a/ptp-tools/Makefile
+++ b/ptp-tools/Makefile
@@ -23,6 +23,7 @@ endif
 KUSTOMIZE ?= $(LOCALBIN)/kustomize
 
 IMG_PREFIX ?= quay.io/<your user id here>/<your image name>
+SAVE_DIR ?= /tmp/ptp-images
 
 VALUES = lptpd cep ptpop krp openvswitch prometheus ptpmg debug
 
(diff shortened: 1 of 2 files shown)


nocturnalastro (approved):
/lgtm

PR #248, opened 2026-06-11
tests: add retries to PTP clock ID lookups

nocturnalastro:
Add Eventually retries at call sites for GetClockIDMaster and GetClockIDForeign to tolerate daemon restart races and reconciliation delays. Scope shared helper retries to pod and profile log ID discovery; leave WaitForClockIDForeign as the explicit long-wait path.

Generated-by: Cursor


jzding (approved):
/lgtm

PR #249, opened 2026-06-12
test: fix DualNICBCHA phc2sys interface selection assertions

nocturnalastro:
When the primary interface recovers from a link-down, ptp4l briefly promotes it to MASTER before a better master is found and it transitions to SLAVE. Each state change (DISABLED->MASTER, MASTER->SLAVE) triggers a phc2sys reconfiguration, emitting a separate "selecting X as out-of-domain source clock" log line. The previous assertions expected exactly one such line in each check window, causing a spurious failure on the recovery check where two lines are produced.

Fix by deduplicating the matched interface names before asserting:
- After primary fails: expect exactly 1 unique interface (the secondary); 2+ unique interfaces indicates flapping.
- After primary recovers: expect 1-2 unique interfaces (secondary during brief MASTER state, then primary on SLAVE transition); 3+ indicates flapping. Take the last unique interface as the settled selection.
- Initial check: take the last match from full pod log history rather than asserting exactly one, making it robust to prior test runs in the same pod session.

Generated-by: Cursor

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -1028,16 +1028,13 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 					fullConfig.DiscoveredClockUnderTestPod.Name, pkg.PtpContainerName,
 					phc2sysLogPattern, false, pkg.TimeoutIn1Minute)
 				Expect(err).NotTo(HaveOccurred())
-				Expect(len(logMatches)).To(BeNumerically("==", 1), "Could not identify which interface phc2sys is using")
-
-				logrus.Infof("phc2sys log matching line: %v", logMatches[0][0])
-				selectedInterface = logMatches[0][1]
+				Expect(len(logMatches)).To(BeNumerically(">=", 1), "Could not identify which interface phc2sys is using")
+				logrus.Infof("phc2sys log matching line: %v", logMatches[len(logMatches)-1][0])
+				selectedInterface = logMatches[len(logMatches)-1][1]
 
 				// Save it as primary interface
 				primaryInterface := selectedInterface
 				By("Verifying the selected interface " + selectedInterface + " is a primary BC's slave interface")
-				// Check if the selected interface belongs to the primary boundary clock config
-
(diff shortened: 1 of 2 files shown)


test/conformance/serial/ptp.go
```diff
@@ -1128,15 +1112,29 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
...
-				Expect(len(logMatches)).To(BeNumerically("==", 1), "Could not identify which interface phc2sys is using")
-				// Get the most recent log entry
-				logrus.Infof("phc2sys log matching line: %v", logMatches[0][0])
-				selectedInterface = logMatches[0][1]
+				Expect(len(logMatches)).To(BeNumerically(">=", 1), "Could not identify which interface phc2sys switched to after primary interface recovered")
+				Expect(countTransitions(logMatches)).To(BeNumerically("<=", 1),
+					fmt.Sprintf("phc2sys made more than 1 interface transition since primary recovered - "+
+						"expected at most 1 (secondary during brief MASTER state, then primary on SLAVE transition); "+
+						"port may be flapping. Sequence: %v",
+						func() []string {
```
edcdavid: Since it is used twice, create a helper function for this

test/conformance/serial/ptp.go
```diff
@@ -1058,20 +1024,32 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 				const phc2sysLogPattern = `phc2sys(?m).*?:.* selecting (\w+) as out-of-domain source clock`
 				var selectedInterface string
 
+				// countTransitions counts the number of times the selected interface changes
+				// between consecutive log entries. Repeated selections of the same interface
+				// (e.g. two bursts both picking ens2f0) count as zero transitions, so only
+				// genuine direction changes increment the counter. This detects flapping
+				// (primary->secondary->primary) while tolerating repeated same-interface lines.
+				countTransitions := func(matches [][]string) int {
```
edcdavid: Why is this function declared inline? Why not create an helper function for this?

edcdavid (approved):

PR #251, opened 2026-06-15
PTP CI: verify PTP processes restart and socket cleanup after termination

yaronh12:
Add serial conformance tests for ptp4l, phc2sys, and ts2phc that verify each process
restarts cleanly after SIGTERM with no event socket errors.

Each test:
- Confirms process_status metric is 1 (running)
- Sends SIGTERM via `pkill`
- Watches for a 1→0→1 metric flip confirming automatic restart
- Scans daemon logs for event socket errors matching the patterns from [OCPBUGS-8112](https://redhat.atlassian.net/browse/OCPBUGS-8112)
The ts2phc test is skipped for non-GM configurations since ts2phc only runs in GM mode.

Jira: [CNF-16780](https://redhat.atlassian.net/browse/CNF-16780)
Regression guard for [OCPBUGS-8112](https://redhat.atlassian.net/browse/OCPBUGS-8112)

## Summary by CodeRabbit

* **Tests**
  * Added coverage for clean restarts of PTP-related processes after a SIGTERM.
  * Verified that restart status metrics flip correctly and that logs stay free of socket-related errors during recovery.
  * Included mode-specific coverage for the additional time-sync process used in GM mode.

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -2239,6 +2239,21 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 
 				logrus.Info("Successfully verified T-BC clock class recovery after upstream link outage")
 			})
+
+			It("Should restart ptp4l with no socket errors after SIGTERM", func() {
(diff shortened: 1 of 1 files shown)


jzding (approved):
/lgm

PR #253, opened 2026-06-15
ci: skip --no-cache for podman builds on GitHub Actions

edcdavid:
## Summary
- GitHub Actions runners start with an empty image store, making `--no-cache` redundant
- Adds a `CACHE_FLAG` variable to `ptp-tools/Makefile` that defaults to `--no-cache` for local builds and is automatically cleared when `GITHUB_ACTIONS` is set
- Reduces build time by allowing layer reuse between parallel builds and avoids `podman save` race conditions caused by concurrent uncached rebuilds

## Test plan
- [ ] Verify Netdevsim CI `ptp-images` job passes without `--no-cache`
- [ ] Verify local `make podman-build-and-saveall` still uses `--no-cache`

## Summary by CodeRabbit

* **Chores**
  * Updated build process configuration to conditionally manage image caching behavior, improving build performance efficiency across different environments.

### DIFF
--- a/ptp-tools/Makefile
+++ b/ptp-tools/Makefile
@@ -25,6 +25,12 @@ KUSTOMIZE ?= $(LOCALBIN)/kustomize
 IMG_PREFIX ?= quay.io/<your user id here>/<your image name>
 SAVE_DIR ?= /tmp/ptp-images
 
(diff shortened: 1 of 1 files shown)


jzding (approved):
/lgtm

PR #259, opened 2026-06-22
ci: use plain podman build in CI instead of manifest lists

edcdavid:
## Summary

- In CI (`GITHUB_ACTIONS` set), use plain `podman build -t` / `podman save` / `podman push` / `podman rmi` instead of manifest-based commands (`podman manifest create` / `build --manifest` / `save -m` / `manifest push` / `manifest rm`).
- CI only builds for a single platform (`linux/amd64`), so manifest lists are unnecessary.
- Fixes a race condition where parallel builds cause `podman save -m` to fail with `image not known` because concurrent manifest storage operations conflict.
- Local (non-CI) builds are unchanged and continue using manifest lists.

## Test plan

- [ ] CI `ptp-images` job passes (previously failing with `Error: could not find image instance ... image not known`)
- [ ] Local `make podman-buildall` still uses manifest-based flow
- [ ] `GITHUB_ACTIONS=true make -n podman-build-debug` shows `podman build -t` (no manifest)
- [ ] `make -n podman-build-debug` shows `podman manifest create` + `podman build --manifest` (unchanged)

## Summary by CodeRabbit

* **Chores**
  * Adapted image build, push, save, and cleanup operations with conditional execution logic for different deployment environments.

### DIFF
--- a/ptp-tools/Makefile
+++ b/ptp-tools/Makefile
@@ -46,14 +46,25 @@ podman-build-%:
 
 podman-pushall: $(foreach v, $(VALUES), podman-push-$(v))
 
+ifdef GITHUB_ACTIONS
+podman-push-%:
+	podman push ${IMG_PREFIX}:$* docker://${IMG_PREFIX}:$*
+else
 podman-push-%:
 	podman manifest push ${IMG_PREFIX}:$* docker://${IMG_PREFIX}:$*
+endif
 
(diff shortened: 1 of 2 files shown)


nocturnalastro (dismissed):
/lgtm

nocturnalastro (approved):
/lgtm

PR #260, opened 2026-06-23
CNF-24031: stop polluting git working tree with generated artifacts

shirmoran:
CNF-24031: stop polluting git working tree with generated artifacts Running run-on-vm.sh left behind artifacts that made git status noisy and confusing after a test run. This commit redirects all generated and downloaded files to a per-run .local-runs/<timestamp> directory so the following stay clean:

-  go.mod / go.sum — removed `go get gomega/...` from install-tools.sh which pulled latest versions and caused dependency drift. Consolidated duplicate go mod tidy/vendor calls into install-tools.sh only.

- vendor/ — no longer modified by redundant go mod vendor in run-on-vm.sh.

- scripts/kind-config.yaml — k8s-start.sh now substitutes registry IP into a temp copy instead of editing the tracked template in-place.

- scripts/openshift-client-linux.tar.gz, scripts/go.tar.gz — tarballs now download to the run directory and are cleaned up after extraction.

- scripts/config.yaml — ginkgo test config now written to the run directory and cleaned up via the existing on_exit trap.

- config/manager/env.yaml, config/manager/kustomization.yaml — the deploy-all Makefile target now runs `git checkout` after kustomize build to restore these files to their committed state.

- scripts/ca.crt — fix-certs.sh no longer writes a temporary ca.crt file; the CA bundle is now fetched directly into a variable.

Run logs (.local-runs/) are now kept by default for post-run debugging. Pass --clean-tmp to delete them automatically.

## Summary by CodeRabbit

* **Chores**
  * Improved setup and test workflows to generate configuration and run artifacts in temporary locations with automatic cleanup (including removing the generated test YAML).
  * Streamlined VM-based execution by isolating per-run logs/artifacts in a timestamped workspace, adding a `--keep-tmp` option to retain it, and simplifying the Go preparation phase.
  * Hardened cluster startup by using a temporary Kind config instead of modifying the tracked file.
  * Updated tool installation to download architecture-specific bundles into temporary locations with reliable cleanup, and simplified Go module preparation.
  * Updated `.gitignore` to exclude per-run temporary artifacts.

### DIFF
--- a/ptp-tools/Makefile
+++ b/ptp-tools/Makefile
@@ -80,9 +80,13 @@ podman-cleanall: $(foreach v, $(VALUES), podman-clean-$(v))
 .PHONY: build-image clean-image deploy-all
 
 deploy-all: 
+	cp ../config/manager/env.yaml ../config/manager/env.yaml.bak
+	cp ../config/manager/kustomization.yaml ../config/manager/kustomization.yaml.bak
 	./scripts/customize-env.sh ${IMG_PREFIX} ../config/manager
 	cd ../config/manager && $(KUSTOMIZE) edit set image controller=${IMG_PREFIX}:ptpop
-	$(KUSTOMIZE) build ../config/default | kubectl apply -f -
+	$(KUSTOMIZE) build ../config/default | kubectl apply -f - ; \
+		mv -f ../config/manager/env.yaml.bak ../config/manager/env.yaml ; \
+		mv -f ../config/manager/kustomization.yaml.bak ../config/manager/kustomization.yaml
 	$(KUSTOMIZE) build ../config/custom | kubectl apply -f -
 	# Operator creates default asynchronously; ensure it exists before patch.
 	kubectl apply -n openshift-ptp -f ../config/samples/ptp_v1_ptpoperatorconfig.yaml
--- a/scripts/install-tools.sh
+++ b/scripts/install-tools.sh
@@ -47,13 +47,18 @@ else
     yum install -y podman pciutils helm
 
     echo "Installing kubectl/oc for $ARCH"
+    _tmp="${PTP_RUN_DIR:-/tmp}"
+    OC_TARBALL="$(mktemp "${_tmp}/openshift-client-linux.XXXXXX.tar.gz")"
     if [[ "$ARCH" == "x86_64" ]]; then
-        curl -Lo ./openshift-client-linux.tar.gz "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz"
+        curl -Lo "${OC_TARBALL}" "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz"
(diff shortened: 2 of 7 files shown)


scripts/install-tools.sh
```diff
@@ -47,13 +47,17 @@ else
     yum install -y podman pciutils helm
 
     echo "Installing kubectl/oc for $ARCH"
+    OC_TARBALL="$(mktemp /tmp/openshift-client-linux.XXXXXX.tar.gz)"
```
edcdavid: Can we store all the temp files in the same per run root unique folder (for instance with timestamp suffix) ? By default the folder is automatically deleted after each run but could be retained with a new option in the run-on-vm.sh script

scripts/run-on-vm.sh
```diff
@@ -14,13 +15,24 @@ while [[ "${1:-}" == --* ]]; do
...
+        --keep-tmp) KEEP_TMP=true; shift ;;
         *) echo "Unknown flag: $1"; exit 1 ;;
     esac
 done
 
-# Save full run output under /tmp/ptp-operator (timestamped file; also shown on the terminal).
-mkdir -p /tmp/ptp-operator
-RUN_ON_VM_LOG="/tmp/ptp-operator/run-on-vm-$(date +%Y%m%d-%H%M%S).log"
+# Per-run temp directory shared by all child scripts.
+export PTP_RUN_DIR="/tmp/ptp-run-$(date +%Y%m%d-%H%M%S)"
```
edcdavid: Looks good to me, just one comment: instead of setting /tmp here which could have a very small size why not create a directory local to the repo that we can add to the .gitignore. For example: .local-runs

edcdavid (commented):
After running the test I saw these remaining file additions/modifications: 
Image

Also, it would be nice to keep the logs from the runs by default instead of deleting at the end

ptp-tools/Makefile
```diff
@@ -113,6 +83,7 @@ deploy-all:
 	./scripts/customize-env.sh ${IMG_PREFIX} ../config/manager
 	cd ../config/manager && $(KUSTOMIZE) edit set image controller=${IMG_PREFIX}:ptpop
 	$(KUSTOMIZE) build ../config/default | kubectl apply -f -
+	git -C .. checkout -- config/manager/env.yaml config/manager/kustomization.yaml
```
edcdavid: We should probably backup these files before modifying them otherwise checkout will loose the user's changes. Maybe we could follow this:
- backup the files
- run the test
- restore the files

edcdavid: We should probably backup these files before modifying them otherwise checkout will loose the user's changes. Maybe we could follow this:
- backup the files
- run the test
- restore the files

edcdavid: We should probably backup these files before modifying them otherwise checkout will loose the user's changes. Maybe we could follow this:
- backup the files
- run the test
- restore the files

edcdavid (dismissed):
looks good to me, pending improvements to the backup./restore of kustomized files (in this PR or another)

edcdavid (approved):
looks good to me !

PR #261, opened 2026-06-23
test: update CI tests for delayed phc2sys startup

lack:
phc2sys is now intentionally delayed in linuxptp-daemon until the clock source (ptp4l or ts2phc) has synchronized. Three test adjustments are needed:

**1. DualNICBCHA:** Before searching for phc2sys log output, explicitly wait for the primary BC slave interface to reach SLAVE state (up to 5 min). phc2sys only starts after ptp4l synchronizes, so the previous 1-minute log search window would time out without this pre-condition.

**2 & 3. "Should run event API v2" (both occurrences):** Increase the cloud-event-proxy `"starting v2 rest api server at port"` log search timeout from 30s to 3 minutes. `WaitForPtpDaemonToBeReady` now returns earlier (phc2sys being delayed no longer blocks the `/ready` endpoint), so cloud-event-proxy has less startup time before the search begins.

**Backwards compatible:** wider timeouts do not affect passing test runs, and the ptp4l SLAVE wait was always a prerequisite for phc2sys to be useful anyway.

Prerequisite for: k8snetworkplumbingwg/linuxptp-daemon#215

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the reliability of conformance checks by waiting for the expected interface state before evaluating log output.
  * Reduced flaky log-matching failures by allowing more time for slower startup paths in API-related checks.
* **Tests**
  * Updated automated test synchronization and timeouts to better handle timing variations in clustered environments.

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -1078,7 +1078,20 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 				logrus.Infof("Primary   BC slave interfaces: %v", primaryBCSlaveInterfaces)
 				logrus.Infof("Secondary BC slave interfaces: %v", secondaryBCSlaveInterfaces)
 
-				// Get phc2sys logs to identify which interface it's using
+				// phc2sys is delayed until ptp4l synchronizes, so first wait for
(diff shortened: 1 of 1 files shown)


lack:
@coderabbit review

nocturnalastro (approved):
/lgtm

PR #263, opened 2026-06-24
CNF-16772: Add CI tests for GM metric cleanup on PtpConfig deletion

yaronh12:
Addresses [CNF-16772](https://redhat.atlassian.net/browse/CNF-16772).

Adds two GM verification tests for metric cleanup on PtpConfig deletion:
1. **OCPBUGS-7808**: Verifies ts2phc clock state metric is removed after GM PtpConfig deletion (stale LOCKED state no longer persists).
2. **OCPBUGS-7811**: Verifies ts2phc config name is preserved (`ts2phc.0.config`) and restart count remains consistent after a GM PtpConfig delete/re-apply cycle.
Also adds helper functions `getClockStateByProcess` and `getProcessConfigAndRestartCount` for parsing PTP metrics.

Passed on CI hardware using `e2e-telco5g-ptp-upstream` with OCP nightly 4.22.

## Summary by CodeRabbit

* **Tests**
  * Added conformance coverage for PTP GM scenarios involving `ts2phc`.
  * Verified clock-state and process metrics are removed when the GM configuration is deleted.
  * Confirmed `ts2phc` recovers correctly after the GM configuration is re-applied, preserving expected config details and restart behavior.

### DIFF
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -2455,6 +2455,112 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
 				}
 			})
 
+			It("Verify ts2phc clock state is not LOCKED after GM PtpConfig is deleted", func() {
+				By("Verifying ts2phc clock state is LOCKED")
+				checkClockStateForProcess(fullConfig, "ts2phc", "1")
+
+				Expect(testconfig.GlobalConfig.DiscoveredGrandMasterPtpConfig).NotTo(BeNil(),
+					"DiscoveredGrandMasterPtpConfig must not be nil")
+				gmConfigName := testconfig.GlobalConfig.DiscoveredGrandMasterPtpConfig.Name
+
+				By(fmt.Sprintf("Deleting GM PtpConfig %s", gmConfigName))
+				err := client.Client.PtpV1Interface.PtpConfigs(pkg.PtpLinuxDaemonNamespace).Delete(
+					context.Background(), gmConfigName, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
+
+				DeferCleanup(func() {
+					By("Recreating GM PtpConfig after test")
+					ensureGMPtpConfigRestored()
+				})
+
+				By("Waiting for ts2phc clock state metric to be removed")
+				Eventually(func() bool {
+					buf, _, _ := pods.ExecCommand(client.Client, true,
+						fullConfig.DiscoveredClockUnderTestPod, pkg.PtpContainerName,
+						[]string{"curl", pkg.MetricsEndPoint})
+					metricsOutput := buf.String()
+					if !strings.Contains(metricsOutput, "# HELP") {
+						return false
+					}
+					_, found := getClockStateByProcess(metricsOutput, "ts2phc")
+					return !found
+				}, pkg.TimeoutIn5Minutes, pkg.Timeout10Seconds).Should(BeTrue(),
+					"Expected ts2phc clock state metric to be removed after GM PtpConfig deletion")
+			})
+
+			It("Verify ts2phc metrics are cleaned up after GM PtpConfig delete and re-apply", func() {
+				Expect(testconfig.GlobalConfig.DiscoveredGrandMasterPtpConfig).NotTo(BeNil(),
+					"DiscoveredGrandMasterPtpConfig must not be nil")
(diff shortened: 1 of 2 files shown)


test/conformance/serial/ptp.go
```diff
@@ -3604,6 +3449,34 @@ func getProcessStatusByProcess(metricsText, process string) (string, bool) {
...
+			if len(parts) == 2 && parts[1] == "1" {
+				start := strings.Index(line, `config="`) + len(`config="`)
+				end := strings.Index(line[start:], `"`)
+				if end > 0 {
+					configName = line[start : start+end]
+					found = true
+				}
+			}
+		}
+		if strings.HasPrefix(line, "openshift_ptp_process_restart_count") && strings.Contains(line, processFilter) {
```
gtannous-spec: would be better putting the second argument as a constant metrics variable: 
`metrics.OpenshiftPtpProcessRestartCount = "openshift_ptp_process_restart_count"`

test/conformance/serial/ptp.go
```diff
@@ -3604,6 +3449,34 @@ func getProcessStatusByProcess(metricsText, process string) (string, bool) {
...
+			parts := strings.Fields(line)
+			if len(parts) == 2 {
+				restartCount, _ = strconv.Atoi(parts[1])
```
gtannous-spec: Maybe should consider adding an error logging for `strconv.Atoi` failure, so malformed metric values don't silently produce zero, while this is unlikely to happen.. but it would be nice to have it as it adds a nice security guard check.. 

(e.g., a float like 3.0), `strconv.Atoi` will silently return 0 and the test will proceed with an incorrect value.


```suggestion
				if len(parts) == 2 {
				var parseErr error
				restartCount, parseErr = strconv.Atoi(parts[1])
				if parseErr != nil {
					fmt.Fprintf(GinkgoWriter, "WARN: could not parse restart_count %q for process %s: %v\n", parts[1], process, parseErr)
				}
			}
```

test/conformance/serial/ptp.go
```diff
@@ -2436,6 +2385,124 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
...
+					"ts2phc did not come back up after GM PtpConfig re-apply")
+				fmt.Fprintf(GinkgoWriter, "After re-apply: ts2phc config=%s, restart_count=%d\n",
+					newConfigName, newRestartCount)
+
+				By("Verifying config name is preserved and metrics are consistent after re-apply")
+				Expect(newConfigName).To(Equal(initialConfigName),
+					fmt.Sprintf("ts2phc config name should be preserved after re-apply, got %s (initial was %s)",
+						newConfigName, initialConfigName))
+				Expect(newRestartCount).To(Equal(initialRestartCount),
+					fmt.Sprintf("ts2phc restart count should be consistent after re-apply, got %d (initial was %d)",
```
gtannous-spec: Do we assume the counter survives a full config (delete+recreation) unchanged ?? @jzding 
(if yes, then ignore this comment). 

However, the restart count assertion `Expect(newRestartCount).To(Equal(initialRestartCount))` assumes that the initial counter remains unchanged.

If the operator resets the metric series on deletion, newRestartCount will be 0 regardless of initialRestartCount that will be causing flaky failures..

test/conformance/serial/ptp.go
```diff
@@ -2436,6 +2385,124 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
...
+				Expect(testconfig.GlobalConfig.DiscoveredGrandMasterPtpConfig).NotTo(BeNil(),
+					"DiscoveredGrandMasterPtpConfig must not be nil")
+				gmConfigName := testconfig.GlobalConfig.DiscoveredGrandMasterPtpConfig.Name
+
+				By(fmt.Sprintf("Deleting GM PtpConfig %s", gmConfigName))
+				err := client.Client.PtpV1Interface.PtpConfigs(pkg.PtpLinuxDaemonNamespace).Delete(
+					context.Background(), gmConfigName, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
+
+				DeferCleanup(func() {
```
gtannous-spec, line 2471: maybe we should consider wrapping this into a helper function as the `defer` logic is duplicated across both tests

jzding (approved):
/lgtm

PR #270, opened 2026-07-13
CNF-25331: fix webhook cert race condition on Kind

shirmoran:
The deployment had a race condition where make deploy-all created the ValidatingWebhookConfiguration before any TLS certificate existed. Until fix-certs.sh manually patched the caBundle, all admission requests failed — blocking linuxptp-daemon DaemonSet creation.

Fix by reordering the deployment and using cert-manager's built-in CA injection:

- Add cert-manager.io/inject-ca-from annotation to the webhook configuration so cert-manager auto-populates caBundle (coexists with the OpenShift annotation for production clusters).

- Apply certs.yaml and wait for Certificates to be Ready before deploying the operator, so the TLS secret is already mounted when the pod starts.

- Remove fix-certs.sh retry loop, sleep, and pod restart — no longer needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment reliability by applying required certificates and waiting for them to become Ready before deploying operator manifests.
  * Added local kind support for webhook CA injection to ensure validating webhook configuration is correctly populated.
  * Streamlined the certificate and webhook initialization sequence to improve consistency across deployments.

### DIFF
--- a/scripts/kind-webhook-ca-patch.yaml
+++ b/scripts/kind-webhook-ca-patch.yaml
@@ -1,0 +1,9 @@
+# Patch for local kind clusters: tells cert-manager to inject the CA bundle
+# from the serving-cert Certificate into the ValidatingWebhookConfiguration.
+# This is NOT needed on OpenShift (service.beta.openshift.io/inject-cabundle handles it).
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: ptpconfig-validating-webhook-configuration
+  annotations:
+    cert-manager.io/inject-ca-from: openshift-ptp/serving-cert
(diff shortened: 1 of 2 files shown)


config/default/webhookcainjection_patch.yaml
```diff
@@ -4,3 +4,4 @@ metadata:
   name: ptpconfig-validating-webhook-configuration
   annotations:
     service.beta.openshift.io/inject-cabundle: "true"
+    cert-manager.io/inject-ca-from: openshift-ptp/serving-cert
```
edcdavid: Create a separate patch in the scripts directory to be used only for local testing with kind cluster

edcdavid (approved):

PR #278, opened 2026-07-30
fix: harden netdevsim CI deploy-prometheus and install-tools

jzding:
## Summary

- **deploy-prometheus.sh**: Replace `podman tag` + `podman push` with `skopeo copy` for creating the `3.4.2` registry tag. Non-CI builds produce manifest lists whose child blobs can be pruned by `podman system prune`, causing "image not known" on push. Skopeo copies directly between registry tags without local image state.

- **install-tools.sh**: Allow `apt-get install` to continue past DKMS autoinstall failures for non-running kernels. Pending kernel header packages trigger netdevsim DKMS builds that fail against newer kernels, blocking the entire install step.

- **netdevsim-ci.yml**: Install `crun` alongside `podman` and print its version during setup to diagnose OCI runtime errors.

## Test plan

- [x] Full `run-on-vm.sh --dkms --deploy` cycle completes without prometheus tag errors
- [x] All 5 test modes pass (oc, bc, dualnicbc, dualnicbcha, dualfollower): 250 tests, 0 failures
- [x] `install-tools.sh` continues past DKMS autoinstall failure for non-running kernel

### DIFF
--- a/scripts/deploy-prometheus.sh
+++ b/scripts/deploy-prometheus.sh
@@ -4,8 +4,12 @@ set -euo pipefail
 
 VM_IP=$1
 
-podman tag "$VM_IP"/test:prometheus "$VM_IP"/test:3.4.2
-podman push  "$VM_IP"/test:3.4.2
+# Create a "3.4.2" registry tag for the prometheus image.
+# `podman tag` + `podman push` doesn't work when the image was built as a
+# manifest list (the default for non-CI builds) because podman may have pruned
+# the underlying image blobs that the manifest index references.
+skopeo copy --src-tls-verify=false --dest-tls-verify=false \
+    "docker://${VM_IP}/test:prometheus" "docker://${VM_IP}/test:3.4.2"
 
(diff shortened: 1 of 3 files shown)


scripts/deploy-prometheus.sh
```diff
@@ -4,8 +4,40 @@ set -euo pipefail
 
 VM_IP=$1
 
-podman tag "$VM_IP"/test:prometheus "$VM_IP"/test:3.4.2
-podman push  "$VM_IP"/test:3.4.2
+# Create a "3.4.2" registry tag for the prometheus image.
```
edcdavid, line 7: what about using skopeo instead:
skopeo copy docker://${VM_IP}/test:prometheus docker://${VM_IP}/test:3.4.2

jzding: Good call — replaced the curl manifest copy with `skopeo copy`. Much cleaner (one command vs the whole curl/python media-type dance), and it handles manifest lists correctly without any of the content-type negotiation issues the bot flagged.

edcdavid (approved):

PR #280, opened 2026-08-03
test: pass PMC command as single arg in getClockIDViaPMC

nocturnalastro:
"GET" and "PARENT_DATA_SET" were passed as separate elements in the exec command slice. Since Kubernetes exec passes each element as a separate argument, pmc treated them as two independent (invalid) commands, producing "bad command: GET" / "bad command: PARENT_DATA_SET".

Combine into a single "GET PARENT_DATA_SET" element to match how pmc expects its command input, consistent with other PMC call sites (e.g. pmc.queryPmc, GetClockClassViaPMC).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the `pmc` command invocation so `GET PARENT_DATA_SET` is processed as a single request.
  * Improved PTP configuration selection to prefer solutions that assign clock roles to distinct nodes, with a fallback when no such solution is available.
  * Applied the improved selection across supported clock configurations, including local-grandmaster Telco BC setups.


edcdavid (dismissed):

jzding (approved):
/lgtm

PR #289, opened 2026-08-19
Enable deploying Cloud Event Proxy v2

sebsoto:
When the EVENT_PROXY_IMAGE env var is set cepv2 will be deployed.

## New Features
- Added optional SideCarV2 event-proxy support for Linux PTP deployments.
- Added configurable image selection for SideCarV2 in operator-managed and generated deployments.
- Preserved the existing event proxy as a fallback when SideCarV2 is not configured.
- Updated event-publishing storage mounts and log forwarding to match the selected event-proxy configuration.

### DIFF
--- a/config/manager/env.yaml
+++ b/config/manager/env.yaml
@@ -19,3 +19,8 @@ spec:
               value: "quay.io/openshift/origin-kube-rbac-proxy:5.0"
             - name: SIDECAR_EVENT_IMAGE
               value: "quay.io/openshift/origin-cloud-event-proxy:5.0"
+            # If this is not empty, this signals to the operator that CEPv2 should be used, and CEPv1 should not be deployed.
+            # The value should be equal to the image location.
+            # The final naming/usage of this should be discussed before release.
+            - name: EVENT_PROXY_IMAGE
+              value: ""
--- a/bindata/linuxptp/ptp-daemon.yaml
+++ b/bindata/linuxptp/ptp-daemon.yaml
@@ -33,6 +33,30 @@ spec:
       priorityClassName: "system-node-critical"
       containers:
         {{ if (eq .EnableEventPublisher true) }}
+        {{ if .SideCarV2 }}
+        - name: cloud-event-proxy
+          image: {{ .SideCarV2 }}
+          imagePullPolicy: {{ .ImagePullPolicy }}
+          command: ["/usr/local/bin/cloud-event-proxy"]
+          args:
+            - "--api-port=9043"
+            - "--store-path=/var/run"
+          volumeMounts:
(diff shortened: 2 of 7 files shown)


config/manager/env.yaml
```diff
@@ -19,3 +19,5 @@ spec:
               value: "quay.io/openshift/origin-kube-rbac-proxy:5.0"
             - name: SIDECAR_EVENT_IMAGE
               value: "quay.io/openshift/origin-cloud-event-proxy:5.0"
+            - name: EVENT_PROXY_IMAGE
```
nocturnalastro, line 25: Are we planning on changing the name or is this a temp name? If its temp I'd like a temp or dev or something like that in the name somewhere to signal it :)

sebsoto: This wont be temporary, but the logic to determine if CEPv1 or CEPv2 is used should be temporary.

Happy to change the name if you have suggestions. I was originally going to go with `SIDECAR_EVENT_V2_IMAGE`, but thats both long, and will make less sense as time goes on, so I changed it to the current state.

nocturnalastro: Do we need to update the Makefile? 

Also we might want to make it so when switch over any SIDECAR_EVENT_IMAGE env var gets set against EVENT_PROXY_IMAGE. We don't want to break any tooling when we switch over. I think the prow CI might use that env var ATM for instance.

I think we need to be careful about unintended consequences from the name change if we switch over to using EVENT_PROXY_IMAGE instead of SIDECAR_EVENT_IMAGE

sebsoto: > Do we need to update the Makefile?

Not at this point, maybe when we switch the operator to using it by default?

And yes, release repo changes will be required, the goal of this was just to be able to get a switch available for CI to be able to use.

sebsoto: > I think we need to be careful about unintended consequences from the name change if we switch over to using EVENT_PROXY_IMAGE instead of SIDECAR_EVENT_IMAGE

I feel using a new variable will make the switch easier, as older branches will still be able to use SIDECAR_EVENT_IMAGE, since this will not be backported.

nocturnalastro: They have there own value of `SIDECAR_EVENT_IMAGE` anyway. I feel like I'm missing something here. Why wouldn't we reuse the name when we switch over? 

Maybe this is a discussion for later as it doesn't actually matter until the switch over, other than we might slightly change this name which I don't feel is that pressing of an issue. 

Perhaps add a comment that we should discuss this in a wider audience before switching over to CEPv2?

nocturnalastro (approved):
/lgtm

PR #299, opened 2026-09-06
CNF-26926: Make linuxptp-daemon verbosity configurable via deployment env

vitus133:
The daemon verbosity is hard-coded in the DaemonSet template ("-v 10"). Introduce a LINUXPTP_VERBOSITY deployment environment variable consumed by the operator and propagated into the rendered linuxptp-daemon container command line.

The value is validated at ingestion (integer 0..14); empty values use the default, and invalid values fall back to the default with a glog warning, so a malformed value never reaches the manifest. PTP protocol behavior (announce/sync intervals, delay mechanism, BMCA) is unchanged.

The variable is declared in the manager manifests and regenerated into the bundle and stable CSVs via make bundle (never hand-edited), so it is visible and editable in production through OLM.

/cc @lack @nocturnalastro @jzding  @sebsoto 

## Summary by CodeRabbit

* **New Features**
  * Added configuration for the LinuxPTP daemon verbosity level through the `LINUXPTP_VERBOSITY` environment variable.
  * Supports verbosity values from 0–14, with a default of 10.
  * Invalid values trigger a warning and use the default verbosity.
* **Bug Fixes**
  * Preserved configured verbosity arguments when updating daemon deployments.


nocturnalastro (approved):
/lgtm

PR #300, opened 2026-09-07
CNF-26944: Add PTP source qualification to PtpClockThreshold

vitus133:
Add ptpSourceQualifiedThreshold (int64 ns), ptpSourceDisqualifiedThreshold (int64 ns), ptpSourceQualifiedSamples / ptpSourceDisqualifiedSamples (int64) and ptpSourceUseS3 (bool, default false)

Assisted by Opencode

## New Features
- Added optional PTP source qualification and disqualification thresholds.
- Added configurable qualification and disqualification sample counts, defaulting to five when unspecified.
- Added an option to use ptp4l servo S3 state for source qualification, defaulting to false.
- Added validation requiring positive values and ensuring the disqualification threshold is at least the qualification threshold.

## Chores
- Updated operator release metadata timestamps.


api/v1/ptpconfig_types.go
```diff
@@ -173,6 +174,25 @@ type PtpClockThreshold struct {
...
+	PtpSourceQualifiedThreshold int64 `json:"ptpSourceQualifiedThreshold,omitempty"`
+	// +kubebuilder:validation:XValidation:rule="self > 0",message="ptpSourceDisqualifiedThreshold must be a positive integer (nanoseconds) when set"
+	// +kubebuilder:default=200
+	// +optional
+	// PTP Source disqualification threshold in nanoseconds. Defaults to 200 ns when omitted.
+	PtpSourceDisqualifiedThreshold int64 `json:"ptpSourceDisqualifiedThreshold,omitempty"`
+	// +kubebuilder:validation:XValidation:rule="self > 0",message="ptpSourceSamples must be a positive integer when set"
+	// +kubebuilder:default=32
+	// +optional
+	// Number of consecutive samples to use for qualification/disqualification (single consolidated counter). Defaults to 32 when omitted.
```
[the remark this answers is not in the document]
vitus133: I changed the ptpSourceSamples to 32 in the PR contract

bundle/manifests/ptp.openshift.io_ptpconfigs.yaml
```diff
@@ -141,6 +141,45 @@ spec:
                               minimum: 0
                               type: integer
                           type: object
+                        ptpSourceDisqualifiedSamples:
```
sebsoto, line 144: Any consideration on adding this to a v2 spec vs the existing config?
I know we want to move that way anyway.

vitus133: Yep, we discussed  `ptpconfigs.ptp.openshift.io/v2` in the context of re-architecting entity relations between Nodes, PtpConfig objects, Profiles and PTP clocks, that should become the center of user intent. I thought `v2` is not in scope for 5.1, while this change is. 
If we feel that `v1` limits our ability to deliver changes we committed to in 5.1 - let's sync and reconsider!

vitus133: Another point @sebsoto - we might need to start backporting these changes all the way down to 4.20. There is a locking time optimization effort ongoing, where one of the vectors could be offset optimizations

